PR #1, opened 2016-02-25
Initial commit with the new build script

victorhurdugaci:
The bootstrapping scripts and the changes required to make it work. No more nuget. We restore with dotnet and the initial bootstrapping is done by downloading the file under `bootstrap`

Will update the `victorhu/init` branch in the script once we're done with the review.

cc @pranavkm @BrennanConroy @natemcmaster


bootstrap/build.cmd
```diff
@@ -0,0 +1,21 @@
+@ECHO off
```
natemcmaster: How about we just make build.cmd launch powershell?

``` cmd
@echo off
powershell -NoProfile -NoLogo -NonInteractive -Command "%~dp0build.ps1" %*
```

natemcmaster:
@victorhurdugaci Is this repo just to split korebuild out of universe?

victorhurdugaci:
Yes, I copied KoreBuild-dotnet here and made my changes on top of it

BrennanConroy:
Looks mostly fine

dougbu:
> I copied KoreBuild-dotnet here and made my changes on top of it

Would be easier to see your changes if you pushed the current Universe files first. Then we could focus on the new stuff.

natemcmaster:
Agree w/ @dougbu. Can't really tell what's changing here.

victorhurdugaci:
> Would be easier to see your changes if you pushed the current Universe files first. Then we could focus on the new stuff.

Nothing else changed except separated the shade files in their own folder, the dotnet install in its own and dnvm in another. 

The changes are what you see in this PR

dougbu:
> No more nuget.

Want to see _exactly_ what has changed here.

victorhurdugaci:
> Want to see exactly what has changed here.

All the changes to remove nuget are in this PR. There are no other changes...

build/shade/_xunit-test.shade
```diff
@@ -32,8 +32,8 @@ configuration=''
         }
 
         var testArgs = noParallelTestProjects.Contains(projectName) ? " -parallel none" : "";
-        var runnerFolder = Path.GetFullPath(Path.Combine(KoreBuildFolderPath, "build", "xunit.runner.console", "tools"));
-        var xunitCoreFolder = Path.GetFullPath(Path.Combine(KoreBuildFolderPath, "build", "xunit.core", "build", "_desktop"));
+        var runnerFolder = Path.GetFullPath(Path.Combine(KoreBuildFolderPath, "xunit.runner.console", "2.1.0", "tools"));
+        var xunitCoreFolder = Path.GetFullPath(Path.Combine(KoreBuildFolderPath, "xunit.core", "2.1.0", "build", "_desktop"));
```
natemcmaster, line 36: Too bad dotnet-restore doesn't support omitting version numbers here.

victorhurdugaci, line 36: Yeah... that's not fun. We could probably find the version programatically but for now this works

template/build.cmd
```diff
@@ -0,0 +1,17 @@
...
+
+SET BOOTSTRAP_FILE_REMOTE=https://raw.githubusercontent.com/aspnet/KoreBuild/victorhu/init/bootstrap/build.cmd
+SET BUILD_FILE=%BUILD_FOLDER%\build.cmd
+
+SET REPO_FOLDER=%~dp0
+CD "%REPO_FOLDER%"
+
+if EXIST %BUILD_FILE% GOTO :EndBootstrapGet
+mkdir %BUILD_FOLDER%
+@powershell -NoProfile -ExecutionPolicy unrestricted -Command "$ProgressPreference = 'SilentlyContinue'; Invoke-WebRequest '%BOOTSTRAP_FILE_REMOTE%' -OutFile '%BUILD_FILE%'"
```
natemcmaster: Same suggestion as above. Why not just do this in powershell and use the 2-line cmd file to launch a ps1 instead? CMD is ugly IMO

bootstrap/build.cmd
```diff
@@ -0,0 +1,21 @@
...
+SET SCRIPT_FOLDER=%~dp0
+SET KOREBUILD_SCRIPT=%SCRIPT_FOLDER%\KoreBuild.cmd
+
+if EXIST "%KOREBUILD_SCRIPT%" GOTO endbootstrap
+
+SET TEMP_CLONE_LOCATION=%TEMP%\KoreBuildClone
+ECHO Downloading KoreBuild in %SCRIPT_FOLDER%...
+
+if EXIST %TEMP_CLONE_LOCATION% RMDIR %TEMP_CLONE_LOCATION% /S /Q
+git clone --branch %KOREBUILD_BRANCH% --depth 1 %KOREBUILD_REPO% %TEMP_CLONE_LOCATION%
```
natemcmaster: Not every windows machine has git in their path. It seems we've replaced nuget.exe with git.exe.

dougbu:
@victorhurdugaci I'm confused by

> ... except separated the shade files in their own folder, the dotnet install in its own and dnvm in another.

versus

> There are no other changes...

Also can't make out how the files in the `bootstrap` or `template` folders are used. Help please.

natemcmaster:
Let me see if I understand the new build step.
1. build.cmd (in repo) downloads (via http) `github.com/..../bootstrap/build.(sh|cmd)`
2. bootstrap/build downloads (via git) `github.com/aspnet/korebuild.git` to `$tempCloneFolder`
3. Finally we call `$tempCloneFolder/korebuild.(cmd|sh)` which 
   1. uses `dotnet-restore` to download sake, xunit, etc.
   2. executes makefile.shade

If thats right, it seems like too many layers of re-direction to get to dotnet-restore, which is what we're really after. Replacing nuget.exe is a weird :chicken: or :egg: problem, but since we still use mono to execute sake anyways, I don't see why we are trying to get off nuget.exe.

template/build.sh
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env bash
+
+buildFolder=.build
+
+bootstrapFileRemote=https://raw.githubusercontent.com/aspnet/KoreBuild/victorhu/init/bootstrap/build.sh
```
dougbu: These URLs are permanent, right? :trollface:

template/build.cmd
```diff
@@ -0,0 +1,17 @@
...
+
+SET BUILD_FOLDER=.build
+
+SET BOOTSTRAP_FILE_REMOTE=https://raw.githubusercontent.com/aspnet/KoreBuild/victorhu/init/bootstrap/build.cmd
+SET BUILD_FILE=%BUILD_FOLDER%\build.cmd
+
+SET REPO_FOLDER=%~dp0
+CD "%REPO_FOLDER%"
+
+if EXIST %BUILD_FILE% GOTO :EndBootstrapGet
```
dougbu: Minor but for simple cases that don't include `SET` commands, an `IF` block is easier to read.

template/build.cmd
```diff
@@ -0,0 +1,17 @@
+@ECHO off
+SETLOCAL
+
+SET BUILD_FOLDER=.build
+
+SET BOOTSTRAP_FILE_REMOTE=https://raw.githubusercontent.com/aspnet/KoreBuild/victorhu/init/bootstrap/build.cmd
+SET BUILD_FILE=%BUILD_FOLDER%\build.cmd
+
+SET REPO_FOLDER=%~dp0
+CD "%REPO_FOLDER%"
```
dougbu: Don't bury the lead. Expect to see these two lines at the top of the file -- just after `SETLOCAL`.

victorhurdugaci:
@natemcmaster nuget.exe is unstable on mono https://github.com/aspnet/Coherence-Signed/issues/178

Your steps are totally correct. The second layer of indirection (bootstrap) is there just so we can make build changes without changing every single repo.

template/build.sh
```diff
@@ -0,0 +1,17 @@
...
+buildFile=$buildFolder/build.sh
+
+repoFolder="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+cd $repoFolder
```
dougbu: Move up to the top.

bootstrap/build.cmd
natemcmaster: Instead, download `https://github.com/aspnet/korebuild/archive/$branch.zip`. Files inside zip are in folder named `KoreBuild-$slugBranch/`.

e.g.Download https://github.com/aspnet/KoreBuild/archive/victorhu/init.zip => unzip => `(unzip)/KoreBuild-victorhu-init/bootstrap/build.cmd`

natemcmaster:
@victorhurdugaci since we can just download KoreBuild as a zip, why not skip the second layer of redirection?
1. In repo, `build.cmd` download github.com/aspnet/korebuild/archive/dev.zip and unzip to `.build`.
2. Execute `.build/korebuild.cmd`

natemcmaster:
Or better yet, we use a git tag. 

`https://github.com/aspnet/korebuild/archive/tag.zip`

This is way you could, in theory have different tags for different repos, if there is a need for slightly different versions of korebuild. (v2? :trollface: ). Multiple tags can point to the same commit too...so updating all repos isn't as big a worry.

bootstrap/build.cmd
```diff
@@ -0,0 +1,21 @@
...
+
+SET KOREBUILD_REPO=https://github.com/aspnet/KoreBuild
+SET KOREBUILD_BRANCH=victorhu/init
+
+SET SCRIPT_FOLDER=%~dp0
+SET KOREBUILD_SCRIPT=%SCRIPT_FOLDER%\KoreBuild.cmd
+
+if EXIST "%KOREBUILD_SCRIPT%" GOTO endbootstrap
+
+SET TEMP_CLONE_LOCATION=%TEMP%\KoreBuildClone
```
dougbu: Multiple builds on same machine shouldn't conflict. Use a unique name (which is easier to do in PowerShell).

build/KoreBuild.cmd
```diff
@@ -47,21 +36,25 @@ IF "%BUILDCMD_DNX_VERSION%"=="" (
...
-    CALL %KOREBUILD_FOLDER%\build\dnvm use default -runtime CLR -arch x86		
+    CALL %KOREBUILD_FOLDER%\dnvm\dnvm use default -runtime CLR -arch x86		
 )		
 REM ============================
 
 :SKIP_RUNTIME_INSTALL
 
+IF EXIST %KOREBUILD_FOLDER%\Sake GOTO :SKIP_DEPS_RESTORE
+dotnet restore %KOREBUILD_FOLDER%\project.json --packages %KOREBUILD_FOLDER%
+:SKIP_DEPS_RESTORE
```
dougbu: Stick w/ `IF` blocks. Odd to have just one `GOTO` in this file.

dougbu:
> why not skip the second layer of redirection?

Like this idea. Doubt we'll need to download the `.zip` file from a different location at some future date.

victorhurdugaci:
- Updated to powershell. The sh files are not yet updated, please don't comment on them.
- Removed the layer of indirection
- Will keep using git because the archive that you get from github has a folder in it and the script would have to figure out which folder is that and the get inside it. Also, without a git repo, it's hard to test locally because you need to create the archive yourself. By using git you can just point to a local folder

natemcmaster:
> Will keep using git because the archive that you get from github has a folder in it and the script would have to figure out which folder is that and the get inside it. Also, without a git repo, it's hard to test locally because you need to create the archive yourself. By using git you can just point to a local folder

Using git as a download client will have its own issues, primarily, that all CI agents and all users need to have Git in their path, which is not always the default option when installing git on windows.

dougbu:
> which is not always the default option when installing git on windows

In addition don't we intend the _entire_ process for building one of our repos to remain `git clone repo` or download and `unzip` the source archive, then `cd repo`, then `.\build.cmd` or `./build.sh`?

victorhurdugaci:
Updated. No more git

build/dotnet/install.sh
```diff
@@ -1,5 +1,3 @@
-#!/usr/bin/env sh
```
victorhurdugaci, line 1: This file is the old `dotnet-install.sh`. Ignore the changes

build/dotnet/install.ps1
```diff
@@ -3,13 +3,21 @@
 # Licensed under the MIT license. See LICENSE file in the project root for full license information.
```
victorhurdugaci, line 3: This file is the old `dotnet-install.ps1`. Ignore the changes

natemcmaster:
LGTM.

build/KoreBuild.ps1
```diff
@@ -0,0 +1,68 @@
...
+$makeFilePath = "makefile.shade"
+if (!(Test-Path $makeFilePath)) 
+{
+    $makeFilePath = "$koreBuildFolder\shade\makefile.shade"
+}
+
+Write-Host "Using makefile: $makeFilePath"
+
+$env:KOREBUILD_FOLDER=$koreBuildFolder
+&"$koreBuildFolder\Sake\0.2.2\tools\Sake.exe" -C $repoFolder -I $koreBuildFolder\shade -f $makeFilePath @args
```
dougbu: Why is `-C $repoFolder` necessary now?

natemcmaster: Sets cwd? just a guess

template/build.ps1
```diff
@@ -0,0 +1,39 @@
...
+{
+    $koreBuildZip=$env:KOREBUILD_ZIP
+}
+
+$buildFolder = ".build"
+$buildFile="$buildFolder\KoreBuild.ps1"
+
+if (!(Test-Path $buildFolder)) {
+    Write-Host "Downloading KoreBuild"    
+    $tempFolder=$env:TEMP + "\KoreBuild-" + [guid]::NewGuid()
```
dougbu: Suggest `[System.IO.Path]::GetRandomFileName()` instead of `[guid]::NewGuid()`. Shorter and still unique enough.

template/build.ps1
```diff
@@ -0,0 +1,39 @@
...
+    
+    if (Test-Path $tempFolder) {
```
dougbu: Don't write code for what should be an impossible case.

template/build.sh
```diff
@@ -0,0 +1,39 @@
...
+
+buildFolder=".build"
+buildFile="$buildFolder/KoreBuild.sh"
+
+if test ! -d $buildFolder; then
+    echo "Downloading KoreBuild"
+    
+    tempFolder="/tmp/KoreBuild-$(uuidgen)"
+    
+    if test ! -d $tempFolder; then
```
dougbu: Shouldn't need this block at all.

template/build.ps1
victorhurdugaci: Leftovers from debugging :)

template/build.sh
```diff
@@ -0,0 +1,39 @@
...
+    
+    wget -O $localZipFile $koreBuildZip 2>/dev/null || curl -o $localZipFile --location $koreBuildZip /dev/null
+    unzip -q -d $tempFolder $localZipFile
+  
+    mkdir $buildFolder
+    cp -r $tempFolder/**/build/** $buildFolder
+    
+    chmod +x $buildFile
+    
+    if test ! -d $tempFolder; then
```
dougbu: `test -d $tempFolder;` since this block should execute if the folder _does_ exist.

build/KoreBuild.ps1
victorhurdugaci: Both of you are correct. Not needed anymore because the script runs in the correct directory now

dougbu:
:shipit: with touch-ups noted above

PR #4, opened 2016-03-01
Updating install.sh script

moozzyk:
The new version does not download a pinned version of the cli if this version is already installed


victorhurdugaci:
I guess you tried it so :shipit:

PR #11, opened 2016-03-03
Downloading nuget.exe when on non-windows

moozzyk:
Fixing a typo in build script

### DIFF
--- a/build/KoreBuild.sh
+++ b/build/KoreBuild.sh
@@ -69,6 +69,12 @@ if [ ! -d $sakeFolder ]; then
     mv "$toolsProject" "$toolsProject.norestore"
 fi
 
+nugetPath="$koreBuildFolder/nuget.exe"
+if [ ! -f $nugetPath ]; then
+    nugetUrl="https://dist.nuget.org/win-x86-commandline/latest/nuget.exe"
+    wget -O $nugetPath $nugetUrl 2>/dev/null || curl -o $nugetPath --location $nugetUrl 2>/dev/null
+fi
+
 makeFile="makefile.shade"
 if [ ! -e $makeFile ]; then
     makeFile="$koreBuildFolder/shade/makefile.shade"
--- a/template/build.sh
+++ b/template/build.sh
@@ -18,7 +18,7 @@ if test ! -d $buildFolder; then
     
     localZipFile="$tempFolder/korebuild.zip"
     
-    wget -O $localZipFile $koreBuildZip 2>/dev/null || curl -o $localZipFile --location $koreBuildZip /dev/null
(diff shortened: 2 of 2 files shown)


moozzyk:
btw. any reason why on Windows we download [nuget only if Sake folder does not exist](https://github.com/aspnet/KoreBuild/blob/dev/build/KoreBuild.ps1#L45)? Not that it is a big deal but I think it these two should be dealt with independently.

build/KoreBuild.sh
```diff
@@ -69,6 +69,12 @@ if [ ! -d $sakeFolder ]; then
     mv "$toolsProject" "$toolsProject.norestore"
 fi
 
+nugetPath="$koreBuildFolder/nuget.exe"
+if [ ! -f $nugetPath ]; then
+    nugetUrl="https://dist.nuget.org/win-x86-commandline/latest/nuget.exe"
```
natemcmaster, line 74: I thought we didn't need to do this anymore....? Don't we use dotnet-restore instead of nuget.exe now?

victorhurdugaci, line 74: We still need nuget to push. It's only for the CI

template/build.sh
```diff
@@ -18,7 +18,7 @@ if test ! -d $buildFolder; then
     
     localZipFile="$tempFolder/korebuild.zip"
     
-    wget -O $localZipFile $koreBuildZip 2>/dev/null || curl -o $localZipFile --location $koreBuildZip /dev/null
+    wget -O $localZipFile $koreBuildZip 2>/dev/null || curl -o $localZipFile --location $koreBuildZip 2>/dev/null
```
victorhurdugaci, line 21: If you do this change then you'll have to update very repo. Are you sure you still want to do it? :smile:

build/KoreBuild.sh
natemcmaster, line 74: Oh, right. Guess there is no dotnet-push yet, eh?

victorhurdugaci, line 74: Exactly

moozzyk, line 74: libuv-build repo needs it to create nuget packages. If we really wanted we might be able to use dotnet for this but it would be very artificial - the package being built contains only native libs and at the moment there is not a single project.json there or even no code since we just pull libuv code as a submodule.

template/build.sh
moozzyk, line 21: I guess we should - otherwise things may not work correctly if you don't have wget

moozzyk, line 21: It does not break really. Even though it shows an error message it will download what we want. Still it would be good to fix this - at the moment `/dev/null` is being treated as an url and you may see this: `curl: (3) <url> malformed`. (Granted almost no one see even this because everyone seems to have wget on their Macs)

victorhurdugaci:
:shipit:

PR #13, opened 2016-03-03
Stop repeatedly adding .NET CLI to `$env:Path`

dougbu:
- environment changes in `KoreBuild.ps1` affect console when running `build.ps1`
  from a PowerShell prompt
- also add .NET CLI to the path even if `$env:KOREBUILD_SKIP_RUNTIME_INSTALL` is set

### DIFF
--- a/build/KoreBuild.ps1
+++ b/build/KoreBuild.ps1
@@ -27,15 +27,21 @@ $dotnetLocalInstallFolderBin = "$dotnetLocalInstallFolder\bin"
 if ($env:KOREBUILD_SKIP_RUNTIME_INSTALL -eq "1") 
 {
     Write-Host "Skipping runtime installation because KOREBUILD_SKIP_RUNTIME_INSTALL = 1"
+    if (!($env:Path.Split(';') -icontains $dotnetLocalInstallFolderBin))
+    {
+        # Add to the _end_ of the path in case preferred .NET CLI is not in the default location.
+        Write-Host "Adding $dotnetLocalInstallFolderBin to PATH"
+        $env:Path = "$env:PATH;$dotnetLocalInstallFolderBin"
+    }
 }
 else
 {
     & "$koreBuildFolder\dotnet\install.ps1" -Channel $dotnetChannel -Version $dotnetVersion
-}
-if (!($env:Path.Split(';') -icontains $dotnetLocalInstallFolderBin))
-{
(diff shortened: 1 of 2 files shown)


victorhurdugaci:
If you set `KOREBUILD_SKIP_RUNTIME_INSTALL` the script shouldn't add the CLI to the path. Or at least it shouldn't add it to the front of the path. 

The purpose of that variable is not to simply skip the installation but it allows you to use a completely different CLI that might exist in a different folder.

victorhurdugaci:
Also, consider changing the `KoreBuild.sh` too

dougbu:
> at least it shouldn't add it to the front of the path

I like this option (add to the end) since `KOREBUILD_SKIP_RUNTIME_INSTALL` could also be used to locally pin the CLI version once `.\build.ps1` downloaded it.

> consider changing the `KoreBuild.sh` too

:ok:

dougbu:
:up::date:

build/KoreBuild.ps1
```diff
@@ -21,18 +21,27 @@ if ($env:KOREBUILD_DOTNET_VERSION)
...
-    $dotnetLocalInstallFolder = "$env:LOCALAPPDATA\Microsoft\dotnet\cli"
-    $dotnetLocalInstallFolderBin = "$dotnetLocalInstallFolder\bin"
     & "$koreBuildFolder\dotnet\install.ps1" -Channel $dotnetChannel -Version $dotnetVersion
-
-    Write-Host "Adding $dotnetLocalInstallFolderBin to PATH"
-    $env:Path = "$dotnetLocalInstallFolderBin;$env:PATH"
+    if (!($env:Path.Split(';') -icontains $dotnetLocalInstallFolderBin))
+    {
+        Write-Host "Adding $dotnetLocalInstallFolderBin to PATH"
+        $env:Path = "$dotnetLocalInstallFolderBin;$env:PATH"
```
victorhurdugaci, line 43: Nit: build the environment value in a string and then just set it once after this block. That way you can also have "Write-Host" in a single place

victorhurdugaci:
After you decide if you want or not to address my comment :shipit:

build/KoreBuild.ps1
dougbu, line 43: :ok:

dougbu:
ab2572a

PR #18, opened 2016-03-06
Embed the commit hash in the generated assembly

victorhurdugaci:
cc @davidfowl @pranavkm

### DIFF
--- a/build/shade/_k-standard-goals.shade
+++ b/build/shade/_k-standard-goals.shade
@@ -97,8 +97,41 @@ default PACKAGELIST_JSON_FILENAME = 'NuGetPackageVerifier.json'
 
 #build-compile target='compile' if='Directory.Exists("src")'
   @{
+    
+    Directory.CreateDirectory(TARGET_DIR);
+    
+    string commitHash = null;
+    if (AddAssemblyInfo)
+    {
+        var commitHashFile = Path.Combine(TARGET_DIR, "commit");
+        GitCommand("rev-parse HEAD >> " + commitHashFile);
+        commitHash = File.ReadAllLines(commitHashFile)[0];
+    }
+  
     var projectFiles = Files.Include("src/*/project.json").ToList();
-    projectFiles.ForEach(projectFile => DotnetPack(projectFile, BUILD_DIR, Configuration));
+    projectFiles.ForEach(projectFile => 
+    {
+        if (AddAssemblyInfo)
+        {
+            var projectText = File.ReadAllText(projectFile);
+            var project = (JsonObject)Json.Deserialize(projectText);
+            var isSharedProject = project.Keys.Contains("shared");
+            
+            // We don't want to embed the commit hash in it because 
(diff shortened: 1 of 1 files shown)


build/shade/_k-standard-goals.shade
```diff
@@ -97,8 +97,35 @@ default PACKAGELIST_JSON_FILENAME = 'NuGetPackageVerifier.json'
...
+  
     var projectFiles = Files.Include("src/*/project.json").ToList();
-    projectFiles.ForEach(projectFile => DotnetPack(projectFile, BUILD_DIR, Configuration));
+    projectFiles.ForEach(projectFile => 
+    {
+        if (AddAssemblyInfo)
+        {
+            var projectFolder = Path.GetDirectoryName(projectFile);
+            var assemblyInfoFile = Path.Combine(projectFolder, "Properties", "AssemblyInfo.cs");
+            if (File.Exists(assemblyInfoFile))
```
Eilon: Why is this check necessary? Shouldn't the SHA always be added?

build/shade/_k-standard-goals.shade
```diff
@@ -97,8 +97,35 @@ default PACKAGELIST_JSON_FILENAME = 'NuGetPackageVerifier.json'
...
+            {
+                var commitHashAttribute = String.Format("[assembly: System.Reflection.AssemblyMetadata(\"CommitHash\", \"{0}\")]", commitHash);
+            
+                var buildInfoFile = Path.Combine(projectFolder, "Properties", "BuildInfo.g.cs");
```
Eilon: Use `.generated.cs` extension?

build/shade/_k-standard-goals.shade
davidfowl: +1 I don't understand what the check is for.

muratg: Do we want to add it for local builds as well?  Let's say I'm building logging... if I haven't made a check in, but trying the build only, I'll get incorrect SHA in there...

build/shade/_k-standard-goals.shade
muratg: +1... that seems to be the standard.

build/shade/_k-standard-goals.shade
Eilon: Not super concerned about local builds having a SHA. I'd rather do "less" in local builds to keep them fast, if possible (that is, avoid running Git).

muratg: So keeping this as an "opt-in" with the environment variable is better? (and not changing the local builds)

victorhurdugaci: Yes, not requiring git on the PATH locally is better, faster, and more reliable. Plus, locally you might have got the code through the git zip file.

victorhurdugaci:
Updated

Eilon:
:shipit:

PR #19, opened 2016-03-06
Use FLOWID if specified when logging messages

pranavkm:


davidfowl:
huh?

pranavkm:
It might make reporting parallel builds a bit easier to read. No idea if it actually works though.

muratg:
Who sets the env var?  Looks like a useful addition to me.

pranavkm:
The Batch build in Universe would make use of it - https://github.com/aspnet/Universe/pull/411/files#diff-9c69ad97a4a478903e84088c363feebaR963

PR #22, opened 2016-03-07
 * Draw version from existing file

ryanbrandenburg:
Make the change to check the correct version for the .sh file too.


ryanbrandenburg:
@BrennanConroy here's the other half of that PR.

BrennanConroy:
:shipit: !

PR #23, opened 2016-03-07
Run xunit tests from build output directory

pranavkm:


build/shade/_xunit-test.shade
```diff
@@ -32,8 +32,7 @@ configuration=''
         }
 
         var testArgs = noParallelTestProjects.Contains(projectName) ? " -parallel none" : "";
-        var runnerFolder = Path.GetFullPath(Path.Combine(KoreBuildFolderPath, "xunit.runner.console", "2.1.0", "tools"));
-        var xunitCoreFolder = Path.GetFullPath(Path.Combine(KoreBuildFolderPath, "xunit.core", "2.1.0", "build", "_desktop"));
```
pranavkm, line 36: This binary gets pulled in via `xunit.extensibility.desktop`

pranavkm:
cc @victorhurdugaci \ @davidfowl

victorhurdugaci:
cc @JunTaoLuo

victorhurdugaci:
It looks okay assuming it doesn't break any tests that look for relative paths and it doesn't break the custom attribute discovery

victorhurdugaci:
Oh, and that it works on mono too

pranavkm:
We can address relative path test (Mvc and Localization are the only ones I know that use it). We can bring back specialized code if it doesn't work on Mono (I can't think of any reasons for it to fail?)

victorhurdugaci:
If it doesn't work on mono, why do you want specialized code?

pranavkm:
_if_ it doesn't work on Mono. I'm not sure it doesn't from the one repo I tried out.

JunTaoLuo:
LGTM, thought I do recall seeing a few more places that use relative paths like Performance. I guess we'll just keep an eye on what fails.

pranavkm:
Cool. It saves a bunch of time in Mvc because publish is never incremental.

muratg:
:shipit: and fix as errors pop up :)

PR #25, opened 2016-03-07
Upgrade the CLI version

victorhurdugaci:
cc @pranavkm @ryanbrandenburg @rynowak


build/cli.version.win
```diff
@@ -1 +1 @@
-1.0.0.001540
\ No newline at end of file
+1.0.0.001651
```
pranavkm: Why is this different?

victorhurdugaci: Because latest windows != latest other today

pranavkm:
:shipit:

rynowak:
Can we hold off on this for like 30 mins? A fix we need is building right now

NTaylorMullen:
Once this is in we should re-transition all of our tools back over to netstandardapp1.5. All should be good supposedly.

pranavkm:
> All she be good supposedly.

Wut?

victorhurdugaci:
> Can we hold off on this for like 30 mins? A fix we need is building right now

Does it require a new CLI?

NTaylorMullen:
> Wut?

It'll also contain fixes for restoring CLI tools that are netstandardapp1.5. We initially transitioned all of our tools to netstandardapp1.5 and then reverted the transition after the CLI was found to not support it.

pranavkm:
:clap:

PR #26, opened 2016-03-08
Execute tests in parallel if --parallel is specified

pranavkm:


pranavkm:
cc @victorhurdugaci \ @BrennanConroy 

Now that we aren't publishing as part of test, we can run them in parallel.

BrennanConroy:
Should we make the number of parallel tests configurable?

Looks good

pranavkm:
I think the Parallel.ForEach scales down (it's `MaxDegreeOfParallelism` and not spawn n-things). We can give it a whirl and see if a setting like that is required at all?

PR #33, opened 2016-03-10
`npm install` shouldn't hide errors and warnings

dougbu:
- `--silent` -> `--quiet`
- add `--depth 0` to get rid of most of the (useless) summary at end
- add `--progress false` when using `--quiet` target; reduces output to _just_ warnings


pranavkm:
:shipit:

dougbu:
c5dcbd7

PR #37, opened 2016-03-16
Rely on dotnet to do incremental builds

pranavkm:
cc @victorhurdugaci \ @javiercn


victorhurdugaci:
:shipit:

pranavkm:
I think we might need to delete any nupkgs that are lying around (since the timestamps keeps changing). I can follow up.

PR #38, opened 2016-03-16
Find properly cased NuGet.config file when using NUGET_VOLATILE_FEED_* variables.

cesarblum:
cc @victorhurdugaci @pranavkm

### DIFF
--- a/build/shade/_use-volatile-feed.shade
+++ b/build/shade/_use-volatile-feed.shade
@@ -1,4 +1,5 @@
 use namespace="System.Collections"
+use namespace="System.IO"
 use namespace="System.Xml.Linq"
 
 @{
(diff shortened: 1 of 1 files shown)


build/shade/_use-volatile-feed.shade
```diff
@@ -10,9 +11,12 @@ use namespace="System.Xml.Linq"
         .Select(entry => new KeyValuePair<string, string>(((string)entry.Key).Substring(prefix.Length), (string)entry.Value))
         .ToList();
 
-    if (feeds.Any())
+    var nugetConfigPath = Directory.EnumerateFiles(Directory.GetCurrentDirectory())
```
pranavkm, line 14: Do you know what repos have files that are differently cased? Perhaps we could fix them instead. dotnet also expects specifying casing.

pranavkm:
Could we address comments before the PR is checked in?

PR #40, opened 2016-03-18
Added an env var to compile and test projects only for the specified framework

ajaybhargavb:
@victorhurdugaci @pranavkm @kichalla

### DIFF
--- a/build/shade/_dotnet-build.shade
+++ b/build/shade/_dotnet-build.shade
@@ -12,10 +12,17 @@ configuration=''
 
 default configuration = 'Debug'
 default build_options=' ${E("KOREBUILD_DOTNET_BUILD_OPTIONS")}'
+default framework = null
 
 @{
     var projectFolder=Path.GetDirectoryName(projectFile);
 
     var dotnetArgs=string.Format("build{0} {1} --configuration {2}", build_options, projectFolder, configuration);
+
+    if (!string.IsNullOrEmpty(framework))
+    {
+        dotnetArgs += " --framework " + framework;
+    }
+
     Dotnet(dotnetArgs);
 }
--- a/build/shade/_k-standard-goals.shade
+++ b/build/shade/_k-standard-goals.shade
@@ -142,13 +142,13 @@ default PACKAGELIST_JSON_FILENAME = 'NuGetPackageVerifier.json'
 #build-test target='compile' if='Directory.Exists("test") && !BuildSrcOnly'
   @{
     var projectFiles = Files.Include("test/*/project.json").ToList();
-    projectFiles.ForEach(projectFile => DotnetBuild(projectFile, Configuration));
+    projectFiles.ForEach(projectFile => DotnetBuild(projectFile, Configuration, BuildFramework));
   }
 
 #build-samples target='compile' if='Directory.Exists("samples") && !BuildSrcOnly'
   @{
     var projectFiles = Files.Include("samples/*/project.json").ToList();
-    projectFiles.ForEach(projectFile => DotnetBuild(projectFile, Configuration));
+    projectFiles.ForEach(projectFile => DotnetBuild(projectFile, Configuration, BuildFramework));
   }
 
 #native-compile target='compile' if='!IsLinux && Directory.Exists(Path.Combine(BASE_DIR, "src"))'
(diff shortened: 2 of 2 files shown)


build/shade/_dotnet-build.shade
```diff
@@ -12,10 +12,17 @@ configuration=''
...
+default framework = null
 
 @{
     var projectFolder=Path.GetDirectoryName(projectFile);
 
     var dotnetArgs=string.Format("build{0} {1} --configuration {2}", build_options, projectFolder, configuration);
+
+    if (!string.IsNullOrEmpty(framework))
+    {
+        dotnetArgs=string.Format("build{0} {1} --configuration {2} --framework {3}", build_options, projectFolder, configuration, framework);
```
kichalla: append to earlier `dotnetArgs` value?

pranavkm: Isn't there an environment variable to pass in options already? Why not use that?

ajaybhargavb: @pranavkm, not sure which one you are referring to.

build/shade/_dotnet-build.shade
```diff
@@ -12,10 +12,17 @@ configuration=''
 
 default configuration = 'Debug'
 default build_options=' ${E("KOREBUILD_DOTNET_BUILD_OPTIONS")}'
```
pranavkm, line 14: This environment variable

build/shade/_k-standard-goals.shade
```diff
@@ -226,18 +227,27 @@ default PACKAGELIST_JSON_FILENAME = 'NuGetPackageVerifier.json'
         var netstandardAppTFM = targetFrameworks.FirstOrDefault(t => t.Equals("netstandardapp1.5", StringComparison.OrdinalIgnoreCase));
         if ((netstandardAppTFM != null || dnxCore50TFM != null) && project.Keys.Contains("testRunner"))
         {
-            DotnetTest(projectFile, Configuration);
+            if (string.IsNullOrEmpty(framework) || framework.Equals(dnxCore50TFM, StringComparison.OrdinalIgnoreCase) || framework.Equals(netstandardAppTFM, StringComparison.OrdinalIgnoreCase))
```
kichalla: You can filter out the `targetFrameworks` to only have the supplied framework and avoid these changes.

build/shade/_dotnet-build.shade
ajaybhargavb, line 14: But it is just easier to have one env var for both build and test

ajaybhargavb:
:up: :date:

build/shade/_k-standard-goals.shade
```diff
@@ -219,6 +219,10 @@ default PACKAGELIST_JSON_FILENAME = 'NuGetPackageVerifier.json'
         var project = (JsonObject)Json.Deserialize(projectText);
         var configs = project.ValueAsJsonObject("frameworks");
         var targetFrameworks = configs == null ? new string[0] : configs.Keys;
+        if (!string.IsNullOrEmpty(BuildFramework))
+        {
+            targetFrameworks = targetFrameworks.Where(t => t.Equals(BuildFramework, StringComparison.OrdinalIgnoreCase)).ToList();
```
victorhurdugaci, line 224: Can you have multiple matches? :laughing:

kichalla, line 224: In what scenario would we have multiple matches?

kichalla, line 224: `ToList` is not required?

ajaybhargavb, line 224: > Can you have multiple matches?
> In what scenario would we have multiple matches?

We won't have multiple matches now. If we want to support more than one framework by having the env var to contain comma separated values. But I don't think it is very useful.

> `ToList` is not required?

`ToList` is required because `targetFrameworks` is of type `ICollection` and `Where` returns `IEnumerable`.

kichalla, line 224: I suppose `.Cast<string>()` would be enough

kichalla:
:shipit:

build/shade/_k-standard-goals.shade
ajaybhargavb, line 224: No, `string` is not the problem. The error will be something like "Cannot assign `IEnumerable<string>` to `ICollection<string>`".

ajaybhargavb:
@pranavkm gave a verbal signoff.

PR #41, opened 2016-03-21
Functions to get typed environment parameters.

cesarblum:
cc @victorhurdugaci

### DIFF
--- a/build/shade/EnvironmentParameters.shade
+++ b/build/shade/EnvironmentParameters.shade
@@ -1,0 +1,15 @@
+use namespace='System'
+ 
+functions @{
+    static string GetEnvironmentParameter(string name)
+    {
+        return GetEnvironmentParameter(name, s => s);
+    } 
+
+    static T GetEnvironmentParameter<T>(string name, Func<string, T> valueFn)
+    {
+        var value = valueFn(Environment.GetEnvironmentVariable(name));
+        Console.WriteLine(name + ": " + (value != null ? value.ToString() : "(null)"));
+        return value;
+    }
+}


build/shade/EnvironmentParameters.shade
```diff
@@ -0,0 +1,20 @@
...
+functions @{
+    static string GetEnvironmentParameter(string name)
+    {
+        return GetEnvironmentParameter(name, s => s);
+    } 
+
+    static T GetEnvironmentParameter<T>(string name, Func<string, T> valueFn)
+    {
+        var value = valueFn(Environment.GetEnvironmentVariable(name));
+        Console.WriteLine(name + ": " + (value != null ? value.ToString() : "(null)"));
```
moozzyk, line 12: what if T is a value type?

build/shade/EnvironmentParameters.shade
```diff
@@ -0,0 +1,20 @@
...
+        return value;
+    }
+
+    static Func<string, string> DefaultTo(string defaultValue)
```
moozzyk: Why would I use this over just `v ?? default`?

One of the concerns I have is that it is hard to find the definition of a Sake function/target and there is no a "go to reference"-like functionality. So if I saw "ToDefault" I would have to find the definition first to understand the script.   Figuring out bugs in Sake scripts is no fun and this in my mind may make it worse.

victorhurdugaci:
:shipit: after you address @moozzyk 's comments

build/shade/EnvironmentParameters.shade
cesarblum, line 12: I hadn't given it proper thought, but it actually works. I'm calling this with a value type (`bool`) to set `skipNpCredentials`.

build/shade/EnvironmentParameters.shade
cesarblum: I'll make the change. I wanted to keep it all within the `GetEnvironmentParameter` functionality but indeed it's a pain to debug Sake scripts.

build/shade/EnvironmentParameters.shade
moozzyk, line 12: I think it depends on what the user does inside the Func because this is where the null will be passed. The only thing is they need to be aware of this possibility. And obviously the message for value types will never hit the path where the `(null)` is printed...

cesarblum, line 12: It's always a string that's passed to the func. The func returns whatever type it wants.

moozzyk:
:ship::it:

PR #44, opened 2016-03-25
Update install.ps1 and install.sh.

NTaylorMullen:
- When we're ready to up our CLI version we need to take this change to enable its retrieval.


victorhurdugaci:
:shipit:

pranavkm:
If this is compatible with the cli we currently use, feel free to check it in.

NTaylorMullen:
It isn't sadly :cry:

PR #48, opened 2016-03-30
Remove dotnet.exe from shared directory to workaround cli issue

pakrym:
https://github.com/dotnet/cli/issues/2143

### DIFF
--- a/build/KoreBuild.ps1
+++ b/build/KoreBuild.ps1
@@ -38,6 +38,9 @@ else
     if ($dotnetCLINew)
     {
         & "$koreBuildFolder\dotnet\install.ps1" -Channel $dotnetChannel -Version $dotnetVersion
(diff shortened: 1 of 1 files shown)


pranavkm:
:shipit: does this fix the issue?

pakrym:
Removing it localy fixed the issue

victorhurdugaci:
:shipit: after you add a comment with a link to the issue or an explanation so that we know why this workaround was added and when to remove it.

davidfowl:
What about the .sh? Is that broken as well?

pakrym:
I don't know exactly there are other blockers on non win so i didn't get to this one.

pranavkm:
Cross plat has the same issues

davidfowl:
Is it being fixed in the same way

PR #49, opened 2016-03-30
Fix dotnet path

pakrym:
New cli install to `"C:\Users\asplab\AppData\Local\Microsoft\dotnet\"` and `dotnet.exe` is directly in there


victorhurdugaci:
:shipit:

PR #50, opened 2016-04-01
Skip logging blocks

pranavkm:
cc @victorhurdugaci 

Not sure if it actually works, but trying to figure out if TeamCity FlowIds work better without blocks.


victorhurdugaci:
Why?

pranavkm:
Didn't work. I'll try this on a separate CI to avoid randomization here.

PR #53, opened 2016-04-05
Add some code to distinguish between linux and darwin

victorhurdugaci:
This will allow us to make the distinction between unix and darwin. It can be used in the CoreCLR build to write the correct file.

cc @pranavkm @ryanbrandenburg

### DIFF
--- a/build/shade/BuildEnv.shade
+++ b/build/shade/BuildEnv.shade
@@ -1,7 +1,17 @@
 use namespace="System"
+use namespace="System.IO"
 
 functions
   @{
+    private static Lazy<bool> _isOSX = new Lazy<bool>(() => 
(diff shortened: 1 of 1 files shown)


ryanbrandenburg:
Seems legit :shipit:

pranavkm:
Could we just hold off until we switch to dotnet-do? We could just use PlatformAbstractions to figure this out instead of doing this.

davidfowl:
YES PLEASE

victorhurdugaci:
Is that still happening?

muratg:
Yes, let's get back to how to do this once RC2 ships.

victorhurdugaci:
Closing this for now

PR #61, opened 2016-04-21
Don't publish symbol packages to SymbolSource

pranavkm:
Publishing to symbol source fails:

```
Pushing NuGetPackageVerifier.1.0.1-rc2-15051.symbols.nupkg to the symbol server (https://nuget.smbsrc.net/)...
  PUT https://nuget.smbsrc.net/api/v2/package/
  Forbidden https://nuget.smbsrc.net/api/v2/package/ 2065ms
Response status code does not indicate success: 403 (Forbidden).
```

### DIFF
--- a/build/shade/_nuget-resilient-publish.shade
+++ b/build/shade/_nuget-resilient-publish.shade
@@ -38,7 +38,7 @@ set nugetArgs='${nugetArgs} ${extra}' if='!string.IsNullOrEmpty(extra)'
         start:
         try
         {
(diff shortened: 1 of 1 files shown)


pranavkm:
cc @victorhurdugaci

victorhurdugaci:
:shipit: I guess... we don't want symbols anymore?

pranavkm:
We never could publish to SybmolSource with the API key we have for MyGet. The new nuget fails for packages with symbol packages next to them, so this should sort it out.

PR #62, opened 2016-04-22
Use build globbing for tests and samples

pakrym:
Should go in only after we get CLI with memory caching.

@pranavkm


pranavkm:
@pakrym does the current CLI support this?

pakrym:
Yes

pakrym:
But I need to make sure before merging that everything works.

pranavkm:
So little faith in your own code :smirk_cat:

pakrym:
Tried on couple repos, seems to work fine.

PR #63, opened 2016-04-27
Temporarily modify dotnet test to specify notecoreapp1.0 as TFM

pranavkm:
cc @pakrym \ @victorhurdugaci

### DIFF
--- a/build/shade/_dotnet-test.shade
+++ b/build/shade/_dotnet-test.shade
@@ -26,7 +26,7 @@ default configuration = 'Debug'
         noParallelTestProjects.UnionWith(NO_PARALLEL_TEST_PROJECTS.Split((char)','));
     }
 
-    var testArgs = " --configuration " + configuration;
+    var testArgs = " -f netcoreapp1.0 --configuration " + configuration;
     testArgs += noParallelTestProjects.Contains(projectName) || IsLinux ? " -parallel none" : "";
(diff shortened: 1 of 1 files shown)


victorhurdugaci:
:shipit:

build/shade/_dotnet-test.shade
```diff
@@ -26,7 +26,7 @@ default configuration = 'Debug'
         noParallelTestProjects.UnionWith(NO_PARALLEL_TEST_PROJECTS.Split((char)','));
     }
 
-    var testArgs = " --configuration " + configuration;
+    var testArgs = " -f netcoreapp1.0 --configuration " + configuration;
```
pranavkm, line 29: The new CLI expects the test runner (`dotnet-test-xunit`) to be a shared dependency because the test command is shared. Options here are 
(a) Change test project.jsons to commonize `dotnet-test-xunit` rather than list it as a `netcoreapp1.0` dependency
(b) Only run for netcoreapp1.0 for now and do `(a)` later (after we've unblocked the gate)

muratg:
Why is this temporary only? When are we going to revert?

pranavkm:
@muratg  We need to switch to dotnet test to run our tests both in all TFMs. That requires project.json across repos

muratg:
Got it. :shipit:

PR #68, opened 2016-05-04
Add flowid to teamcity messages to improve parallel building experience

pakrym:
Universe mostly
@pranavkm @victorhurdugaci


pranavkm:
:shipit:

pranavkm:
Does it actually work?

pakrym:
Docs say it does

pakrym:
It doesn't work, reverted

PR #79, opened 2016-05-23
Ignore dotnet-test-xunit exit code if KOREBUILD_IGNORE_DOTNET_TEST_EXIT_CODE is set

### LINKED ISSUE #78: Ignore non-zero exit from dotnet-test-xunit when running under TeamCity
The xunit runner is supposed to return a 0-exit code when one or more tests fail when running under TeamCity. TeamCity will report these errors and fail the build. However it returns a non-zero exit code. This prevents features such as test muting from working. 

The fix here would be to ignore the exit code in `_dotnet-test.shade` when running in TeamCity.

pranavkm:
Fixes #78

cc @cesarbs \ @troydai

### DIFF
--- a/build/shade/_dotnet-test.shade
+++ b/build/shade/_dotnet-test.shade
@@ -15,6 +15,7 @@ configuration=''
 */}
 
 default configuration = 'Debug'
+default IgnoreDotnetTestExitCode = '${E("KOREBUILD_IGNORE_DOTNET_TEST_EXIT_CODE") == "true"}'
 
 @{
     var content = File.ReadAllText(projectFile);
@@ -37,7 +38,18 @@ default configuration = 'Debug'
         }
 
         testArgs += noParallelTestProjects.Contains(projectName) ? " -parallel none" : "";
-        Dotnet("test" + testArgs, projectFolder);
+
+        try
+        {
+            Dotnet("test" + testArgs, projectFolder);
+        }
+        catch
+        {
+            if (IgnoreDotnetTestExitCode)
+            {
+                // Skip dotnet-test exit code.
+            }
+        }
     }
 }
 


build/shade/_dotnet-test.shade
```diff
@@ -37,7 +38,18 @@ default configuration = 'Debug'
         }
 
         testArgs += noParallelTestProjects.Contains(projectName) ? " -parallel none" : "";
-        Dotnet("test" + testArgs, projectFolder);
+
+        try
+        {
+            Dotnet("test" + testArgs, projectFolder);
+        }
+        catch
```
troydai, line 46: could this accidentally swallow any exception which really should have failed the test?

pranavkm, line 46: There's a risk of it, but there's not much we could do short of matching the exception message here: https://github.com/sakeproject/sake/blob/master/src/Sake.Library/Shared/_exec.shade#L60

natemcmaster:
Why does dotnet-test-xunit return non-zero?

pranavkm:
@natemcmaster it seems like a bug in the runner. This was stop gap to allow us to mute tests until this is fixed.

natemcmaster:
Is there a bug on dotnet-test-xunit? Under what conditions do tests pass and it still returns a non-zero exit code?

pranavkm:
@natemcmaster it's the other way around. The tests fail and TeamCity is aware of it, but the test runner fails with a non-zero exit code which causes the build to fail. Muting doesn't help in this case.

natemcmaster:
Thanks for clarifying. Could we key this behavior off the "IfTeamCity" variable instead?

pranavkm:
I didn't want to turn it on the CoreCLR build in case there was something that broke the runner.

PR #81, opened 2016-05-31
Don't ignore dotnet-test exit code if enviornment isn't set

pranavkm:
cc @divega


divega:
:shipit:

divega:
And thanks for the quick fix.

PR #82, opened 2016-06-07
Initial work on "KoreBuild 2.0"

### LINKED ISSUE #83: Add support for a side-by-side "KoreBuild 2.0"
Now that MSBuild is going cross-platform and we're transitioning other things to it, we should move KoreBuild to MSBuild and away from the custom tools we're using right now (i.e. Sake).

In order to start this process, we should create a KoreBuild 2.0 where we can start adding this new build orchestration and set KoreBuild up so that each repo can use either of KoreBuild 1.0 or 2.0, in order to allow an incremental transition.

(/cc @muratg @Eilon This is part of the MSBuildification work I've been doing in between code reviewing/RTM work :smile:. Targeting 1.0.1)

analogrelay:
- Set up a side-by-side system to allow repositories to use either KoreBuild 1.0 or KoreBuild 2.0
- Prepare a new MSBuild-based KoreBuild to be used after ASP.NET Core 1.0 RTM
- Set up an infrastructure for easy development of custom MSBuild tasks
- Create a script to install KoreBuild (2.0) into a repository

**IMPORTANT Stuff:**
- This does **NOT** actually contain any of the implementation of KoreBuild 2.0's MSBuild targets. Just the logic to bootstrap MSBuild and the KoreBuild tasks. MSBuild angle brackets will come in a later PR. If you're curious, see the [`anurse/msbuild-targets` branch](https://github.com/aspnet/KoreBuild/commits/anurse/msbuild-targets) in this repo.
- This change is fully backwards-compatible, since it is entirely parallel to the existing system. Repositories that have not been converted (using `install.ps1` or `install.sh`) will continue to use the existing system (I have verified this)
- The PR currently has hard-coded defaults to the `anurse/msbuild` branch. I'll change this immediately prior to merging. For now I want to leave them in to ensure the PR works as-is.

You can test this if you'd like, in any repo:
- To test backwards-compat, open the repo's `build.ps1` or `build.sh` and change the URL of the KoreBuild ZIP from `https://github.com/aspnet/KoreBuild/archive/dev.zip` to `https://github.com/aspnet/KoreBuild/archive/anurse/msbuild.zip`. This will force the old KoreBuild scripts to use the new branch, proving that it still works
- To test-convert a repo to the new thing, clone KoreBuild somewhere and check out this branch (`anurse/msbuild`), cd to the repo you want to convert and run the `install.sh`/`install.ps1` from the KoreBuild repo in that directory. Then just `build.sh` or `build.ps1` in the repo :). You should get the "KoreBuild 2.0 is not ready" error.

**This change targets POST 1.0, so I'm going to hold it until after we create 1.0 release branches**

Fixes #83 

/cc @davidfowl @victorhurdugaci @pranavkm @troydai 
/fyi @muratg @Eilon


build2/KoreBuild.ps1
```diff
@@ -0,0 +1,100 @@
+Write-Host -ForegroundColor Green "Starting KoreBuild 2.0 ..."
+
+$RID = "win7-x64";
+if($env:PROCESSOR_ARCHITECTURE -eq "x86") {
+    $RID = "win7-x86";
+}
+
+$ToolsDirectory = $env:KOREBUILD_TOOLS_DIRECTORY
+if(!$ToolsDirectory) {
+    $ToolsDirectory = Join-Path $env:LOCALAPPDATA "Microsoft\AspNetCoreBuildTools\$RID"
```
victorhurdugaci: This might not be the best solution even if configurable because then you can have only one version of the build system and if you build a different repo you might update the build scripts for everyone. Also, doing multiple builds on the same machine at the same time might crash

install.ps1
```diff
@@ -0,0 +1,80 @@
+# Installs KoreBuild into a repo, overwriting all previous build scripts
+# This script is downloaded and executed without any of the corresponding scripts in this repo, so don't assume anything in this repo is available!
+
+param(
+    [string]$KoreBuildRepo,
+    [string]$KoreBuildBranch = "anurse/msbuild") # TEMPORARY until this is merged to dev
+
+# Prerequisites
+if(!(Get-Command git -ErrorAction SilentlyContinue)) {
+    throw "KoreBuild requires Git"
```
victorhurdugaci: Oh, this is bad. We tried as much as possible to not require Git in KoreBuild. The EF CI machines don't have git

install.ps1
```diff
@@ -0,0 +1,80 @@
...
+}
+if(!(Test-Path "$(Get-Location)\.git")) {
+    throw "This install script should be run from the root of a Git repo"
```
victorhurdugaci, line 38: What if I download the code as zip?

install.ps1
```diff
@@ -0,0 +1,80 @@
...
+}
+$Changes = git diff --name-only
+if($Changes) {
+    throw "There are unstaged changes in this repo. Reset or commit them before installing KoreBuild"
```
victorhurdugaci: Why?

template2/build.sh
```diff
@@ -0,0 +1,90 @@
...
+        --korebuild-repo)
+            KOREBUILD_REPO=$2
+            shift
+            ;;
+        --korebuild-branch)
+            KOREBUILD_BRANCH=$2
+            shift
+            ;;
+        --korebuild-dest)
+            KOREBUILD_DEST=$3
```
victorhurdugaci: Why `$3`?

install.ps1
analogrelay, line 38: Then you can't **install** KoreBuild into it. If you download a repo that already has KoreBuild this script isn't ever used.

install.ps1
analogrelay: a) Hygiene. Don't install KoreBuild as part of a bigger commit
b) Ensures you can easily see what the install script did

I don't really care, we can remove it, it just seemed like a resonable restriction to place on the auto-installer for KoreBuild. If you want to do otherwise, you can always manually install it.

install.ps1
analogrelay: Here, we're talking about **installation**. CIs shouldn't be installing KoreBuild into repos, only using it.

We do use `git` in the build script itself to ensure KoreBuild is always up-to-date. I think the benefits of that (never having to delete `.build` to force an update) outweigh the costs of installing git on a couple of machines, but I'm certainly open to discussion.

build2/KoreBuild.ps1
analogrelay: Agreed, this is a short-term solution. I'm not sure the best option here. I may take another look at having it installed in each repo and see what the impact of that is. The big problem is that we need to do a full restore of all the CoreCLR binaries **and** publish them to make sure CoreCLR works.

template2/build.sh
analogrelay: Because shut up, that's why :P

build2/KoreBuild.ps1
analogrelay: The tools alone end up taking up 50MB in the repo if we do it per-repo... At this point I'm leaning towards saying that's still better than having it in a central location, but it is definitely kinda gross. I'll set it back to using a repo-local install for now, but we'll need to look at solutions for shrinking it.

We could consider making MSBuild Core a pre-req and assuming you've got it somewhere (and providing a script to install it if you don't).

install.ps1
natemcmaster: We've discussed this many times before. Although we like the benefit of auto-updating build scripts, there are better ways to accomplish this than using git IMO.

build2/KoreBuild.ps1
analogrelay: FYI We already do this for the CLI btw, it is installed to a global location during build. We should look at reconsidering that too.

install.ps1
analogrelay: What's the concern around installing git on the CI machines? Also, I'm not sure there are better ways to accomplish it, but I'm certainly open to options :). I'd rather not invent something new when we're already using something that works well.

natemcmaster: > we're already using something that works well.

Ay there's the rub. We're not using git yet to synchronize korebuild version. We don't know all issues we'll hit once we start using it. But here are some things to consider
- `git clean -xfd` doesn't delete sub-git folders. People have to re-train muscle memory to add the second "f" to actually clean everything `git clean -xffd`
- `git pull` isn't enough. Homebrew has [extensive code](https://github.com/Homebrew/brew/blob/master/Library/Homebrew/cmd/update.sh) around `git pull` to ensure that if a simple pull fails, it can recover. We'll need something similar otherwise the standard fix will rename "just delete `.build`".
- Switching branches. git checkout will not always work if there will be file conflicts as a result of the branch change.

I'm not opposed to using git per se ... but I assert there are simpler cache invalidation strategies.

analogrelay: Alrighty, you convinced me :). I'm not sure these aren't issues we'd have with our own auto-update mechanism, but I'm convinced that it's not worth having an auto-update mechanism given the issues :D

build2/KoreBuild.ps1
```diff
@@ -0,0 +1,100 @@
...
+    [IO.File]::WriteAllText((Join-Path $MSBuildProjDir "project.json"), $content);
+
+    copy "$PSScriptRoot\NuGet.config" "$MSBuildProjDir"
+
+    Write-Host -ForegroundColor Green "Preparing build tools..."
+
+    exec dotnet restore "$MSBuildProjDir\project.json" -v Detailed
+    exec dotnet publish "$MSBuildProjDir\project.json" -o "$MSBuildDir"
+    exec dotnet restore "$PSScriptRoot\.." -v Detailed
+    exec dotnet publish "$PSScriptRoot\..\src\Microsoft.AspNetCore.Build.Tasks" -o "$MSBuildDir" -f "netcoreapp1.0"
```
pranavkm: Do we know what kind of tasks we plan on adding here? The things we usually run (build, compile, pack and test) in most repos can be expressed as vanilla msbuild targets

install.ps1
```diff
@@ -0,0 +1,108 @@
+# Installs KoreBuild into a repo, overwriting all previous build scripts
```
pranavkm, line 1: We're probably going to bulk update repos as a one time (basically how @victorhurdugaci updated to the current KoreBuild). Wouldn't we just do this offline (without having to commit this)?

build2/KoreBuild.ps1
analogrelay: In my other branch where I'm actually building the MSBuild targets, I tried to avoid it but did end up needing custom tasks. Primarily to do things like reading `project.json` to determine if a test project targets `net451`

install.ps1
analogrelay, line 1: Sure, but it also helps for new repos, etc. and it doesn't hurt to have the script we're going to use committed.

I do think we probably only need one of the ps1/sh though...

build2/KoreBuild.ps1
analogrelay: See https://github.com/aspnet/KoreBuild/tree/anurse/msbuild-targets/src/Microsoft.AspNetCore.Build.Tasks

analogrelay:
Ping. I've updated the scripts to avoid using Git now. I also removed the `install.sh` since the "install" script is something that only needs to be run once and I don't want to risk bit-rot from having two different kinds of install script. It means you can't auto-install KoreBuild into a repo from a Mac/Linux machine, but I think we can handle that. The manual install isn't exactly hard (copy `build.*` from `template2` to the root of the repo)

build2/KoreBuild.sh
```diff
@@ -0,0 +1,94 @@
...
+        echo "Skipping runtime installation because KOREBUILD_SKIP_RUNTIME_INSTALL is set"
+
+        # Add .NET installation directory to the path if it isn't yet included.
+        # Add to the _end_ in case preferred .NET CLI is not in the default location.
+        [[ ":$PATH:" != *":$DOTNET_INSTALL_DIR:"* ]] && export PATH="$PATH:$DOTNET_INSTALL_DIR"
+    else
+        # Need to set this variable because by default the install script
+        # requires sudo
+        export DOTNET_INSTALL_DIR=~/.dotnet
+        export KOREBUILD_FOLDER="$(dirname $DIR)"
```
victorhurdugaci: The `ps1` script doesn't set this and there are some builds that depend on it

template2/build.ps1
```diff
@@ -0,0 +1,97 @@
+# KoreBuild 2.0
```
victorhurdugaci, line 1: The old script used to set `REPO_FOLDER`. I think there are some builds that depend on it

build2/KoreBuild.ps1
```diff
@@ -0,0 +1,100 @@
...
+    Write-Host -ForegroundColor Cyan "> $cmd $args"
+    "`r`n>>>>> $cmd $args <<<<<`r`n" >> $KoreBuildLog
+    & $cmd @args >> $KoreBuildLog
+    if($LASTEXITCODE -ne 0) {
+        throw "Command returned exit code $($LASTEXITCODE): '$cmd $args'"
+    }
+}
+
+function EnsureDotNet() {
+    $dotnetVersionFile = $PSScriptRoot + "\..\build\cli.version.win"
```
victorhurdugaci: If you set `KOREBUILD_FOLDER` then you can use it here

build2/KoreBuild.ps1
```diff
@@ -0,0 +1,100 @@
...
+# Check for a local KoreBuild project
+$Proj = Join-Path (Get-Location) "makefile.proj"
+if(!(Test-Path $Proj)) {
+    $Proj = Join-Path "$PSScriptRoot" "msbuild\makefile.proj"
+}
+
+$LogFile = Join-Path $PSScriptRoot "..\korebuild.msbuild.log"
+
+Write-Host -ForegroundColor Cyan "> msbuild $Proj $args"
+& "$MSBuildDir\CoreRun.exe" "$MSBuildDir\MSBuild.exe" /nologo $Proj /p:KoreBuildTasksPath="$MSBuildDir" /fl "/flp:logFile=$LogFile;verbosity=diagnostic" @args
```
victorhurdugaci: Quotes around `$Proj`. (and any other place where you pass paths to an executable

template2/build.sh
```diff
@@ -0,0 +1,97 @@
...
+    # Sub-shell (to preserve current dir)
+    mkdir $BUILD_FOLDER
+    cd $BUILD_FOLDER
+    tar xf $KOREBUILD_TAR --strip-components 1
+    rm -Rf $KOREBUILD_DIR
+fi
+
+cd $DIR
+chmod a+x $BUILD_FILE
+$BUILD_FILE "$@"
```
victorhurdugaci, line 115: Will this pass all the arguments, including those passed to this script (e.g `--reset-korebuild`) further?

analogrelay, line 115: No, I tested it out and it will pass each argument from the current `$1` on, so the `shift` calls above remove the leading KoreBuild.sh arguments from that list.

victorhurdugaci, line 115: Ah,  nice! I also missed the `break`s in that switch

build2/KoreBuild.ps1
analogrelay: PowerShell is _supposed_ to handle that when you only pass a single variable value but I agree paranoia is the best policy there :)

build2/KoreBuild.sh
analogrelay: Interestingly, the original `KoreBuild.sh` script sets it twice and to [different](https://github.com/aspnet/KoreBuild/blob/d7473ac46c470112fbd9e4d5fe3fe93f2d4305d9/build/KoreBuild.sh#L54) [values](https://github.com/aspnet/KoreBuild/blob/d7473ac46c470112fbd9e4d5fe3fe93f2d4305d9/build/KoreBuild.sh#L95) so any build that does depend on it is going to be a bit wonky.

analogrelay:
Barring further comments, I'm hoping to merge this shortly after we branch `release` so we can start iterating on it and work on the actual build implementation.

victorhurdugaci:
We should rename this to MorkBuild

analogrelay:
Microsoft ASP.NET Core Build Tools for Engineers 2015

victorhurdugaci:
SP1

analogrelay:
I'm going to make one final modification: I'm moving the directory KoreBuild is extract to to `.build\KoreBuild` and then it will set up the tools (MSBuild, Tasks, NuGetPackageVerifier, etc.) in `.build\Tools`. This has two advantages: a) It makes KoreBuild immutable, so while developing KoreBuild you can run from a different path and b) It allows us to easily detect if the tools have already been prepared so we don't run a bunch of `restore` and `publish` ops on every build.

build2/KoreBuild.ps1
```diff
@@ -0,0 +1,104 @@
...
+$ToolsDirectory = $env:KOREBUILD_TOOLS_DIRECTORY
+if(!$ToolsDirectory) {
+    $ToolsDirectory = Join-Path $KoreBuildScriptDir "Tools"
+}
+
+$MSBuildProjDir = Join-Path $ToolsDirectory "MSBuildTools"
+$MSBuildDir = Join-Path $MSBuildProjDir "publish"
+$KoreBuildTasksPath = Join-Path $ToolsDirectory "KoreBuild"
+
+$KoreBuildLog = Join-Path $KoreBuildScriptDir "..\korebuild.log"
```
troydai: nit: save it to a log folder and put a timestamp on it? i find it usefully to save logs for comparison to detect issues. - non block suggestion.

global.json
```diff
@@ -0,0 +1,3 @@
+﻿{
+  "projects": [ "src", "test" ]
```
troydai: is there a test folder?

troydai:
LGTM

build2/KoreBuild.ps1
analogrelay: I don't want to fill the log folder up, nor to I want to create a whole log rotation thing. I think we'll leave that for a later iteration.

global.json
analogrelay: Nope, just part of the template.

src/Microsoft.AspNetCore.Build.Tasks/project.json
```diff
@@ -0,0 +1,26 @@
+﻿{
+    "version": "1.0.0-*",
```
pranavkm: 2 spaced indentation.

pranavkm:
:shipit:

src/Microsoft.AspNetCore.Build.Tasks/project.json
analogrelay:
```
     O       K
```

Spaces

davidfowl:
👏

PR #87, opened 2016-06-14
Revert "Revert "Revert "Make NuGet packages serviceable"""

natemcmaster:
@pranavkm 

Until nuget.exe 3.5.0-beta2 is available, let's instead set this option via KOREBUILD_DOTNET_PACK_OPTIONS on CI.


pranavkm:
I'd prefer if we added a new build target `ci-build` that sets the environment variables up rather than relying on the CI to do this.

natemcmaster:
Close in favor of #88

PR #94, opened 2016-06-28
Restore reference assemblies in xplat

pranavkm:
cc @victorhurdugaci \ @pakrym \ @anurse

### DIFF
--- a/build/KoreBuild.ps1
+++ b/build/KoreBuild.ps1
@@ -62,6 +62,13 @@ if (!(Test-Path "$koreBuildFolder\Sake"))
     &dotnet restore "$toolsProject" --packages "$PSScriptRoot" -v Minimal
     # Rename the project after restore because we don't want it to be restore afterwards
     mv "$toolsProject" "$toolsProject.norestore"
+
+    $xplatRestoreProject = "$koreBuildFolder/xplat.project.json"
+    if (Test-Path $xplatRestoreProject)
+    {
+        mv $xplatRestoreProject "$xplatRestoreProject.norestore"
+    }
+
     # We still nuget because dotnet doesn't have support for pushing packages
     Invoke-WebRequest "https://dist.nuget.org/win-x86-commandline/latest/nuget.exe" -OutFile "$koreBuildFolder/nuget.exe"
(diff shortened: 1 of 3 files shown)


pranavkm:
This is required for https://github.com/aspnet/Logging/pull/448

pakrym:
Can we call the package `NETFrameworkReferenceAssemblies` to be a bit more specific?
Otherwise :shipit:

natemcmaster:
Just curious. Have we looked into using the official targeting pack nupkgs? 

https://dotnet.myget.org/feed/dotnet-core/package/nuget/Microsoft.TargetingPack.NETFramework.v4.5.1

pranavkm:
Oh nifty - didn't realize they had a package we could target. I'll give that one a try

natemcmaster:
Actually, I just tried it on Microsoft.Data.Sqlite. It doesn't seem to work :-/

pranavkm:
@natemcmaster could you file a CoreFx bug for it so that we can eventually move to their package.

natemcmaster:
https://github.com/dotnet/corefx/issues/9735

pakrym:
There is no CLI support for this. We still can use this package same way we used our own.

pranavkm:
Yeah, looks like it. In our case, we don't reference the package but treat it as the search space for reference assemblies. Maybe doing something similar works just fine?

pranavkm:
Their structure is kinda crummy - `/Microsoft.TargetingPack.NETFramework.v4.5.1/1.0.1/lib/net451` has binaries in it whereas `DOTNET_REFERENCE_ASSEMBLIES_PATH` maps to `C:\Program Files (x86)\Reference Assemblies\Microsoft\Framework`. We'd have to move `net451` -> `.NETFramework/v.4.5.1` something our package already preserves. @natemcmaster do you feel strongly about switching to this package? If not we can live with what we have which seems to be working fine so far.

natemcmaster:
No, it looks the like the packages aren't really ready for use anywhere than in corefx's build system. They're kinda working on making it possible, but it's not ready yet.

PR #96, opened 2016-06-29
Add MSBuild code to start replacing KoreBuild.

analogrelay:
Ok, here we go. The first pass at MSBuilding stuff. This MSBuild script can basically successfully build aspnet/Options. It produces the NuGet packages and in theory it has the necessary stuff to push to MyGet (though that's likely to need iterating).

&lt;I for="one"&gt;welcome&lt;!--our--&gt;new &amp;XML;&lt;/overlords&gt;

/cc @pranavkm @victorhurdugaci @javiercn @davidfowl


src/Microsoft.AspNetCore.Build/NuGetInstall.cs
```diff
@@ -0,0 +1,68 @@
...
+                {
+                    Log.LogError($"Package file not found: {path}");
+                    success = false;
+                }
+                else
+                {
+                    using (var stream = File.OpenRead(path))
+                    {
+                        PackageIdentity id;
+                        using (var reader = new PackageArchiveReader(stream, leaveStreamOpen: true))
```
pranavkm: I wonder if this would be problematic when they have breaking changes for e.g. when they updated the schema version so that you had to reference their nightlies to read packages. We've referenced their feeds in these cases to workaround it in the past.

analogrelay: It may be... I do think we should always be referencing a pretty bleeding-edge NuGet to help avoid that. If we end up having to use our own code rather than the APIs we could do that too.

src/Microsoft.AspNetCore.Build/DetectOSPlatform.cs
```diff
@@ -0,0 +1,33 @@
+﻿using System.Runtime.InteropServices;
```
victorhurdugaci: Copyright?

src/Microsoft.AspNetCore.Build/DetectOSPlatform.cs
```diff
@@ -0,0 +1,33 @@
...
+            {
+                PlatformName = "Windows";
+            }
+            else if (RuntimeInformation.IsOSPlatform(OSPlatform.Linux))
+            {
+                PlatformName = "Linux";
+            }
+            else if (RuntimeInformation.IsOSPlatform(OSPlatform.OSX))
+            {
+                PlatformName = "OSX";
```
victorhurdugaci: MacOS :trollface:

src/Microsoft.AspNetCore.Build/NuGetResilientPush.cs
```diff
@@ -0,0 +1,127 @@
...
+                        path,
+                        Feed,
+                        ApiKey,
+                        TimeoutInSeconds,
+                        disableBuffering: false,
+                        noSymbols: false,
+                        logger: logger);
+                    pushed = true;
+                }
+                catch (Exception) when (tries < Retries)
```
victorhurdugaci: Nice!

src/Microsoft.AspNetCore.Build/targets/Package.targets
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="14.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+    <!-- Put targets attached to the 'Clean' phase of the Standard Lifecycle here. -->
+
+    <Target Name="CleanPreviousPackOutput">
+        <RemoveDir Directories="$(PackagesDir)" />
+    </Target>
+    <Target Name="DotNetPack" DependsOnTargets="CleanPreviousPackOutput" BeforeTargets="Package">
+        <Exec
+            Command="dotnet pack --version-suffix $(BuildVersionSuffix) -o &quot;$(PackagesDir)&quot; --no-build --configuration $(Configuration) $(DotNetPack_Options) %(Projects.ProjectDir)"
```
victorhurdugaci: quote all path arguments

src/Microsoft.AspNetCore.Build/DetectOSPlatform.cs
analogrelay: Copyright 2016 Andrew! Take that Microsoft!

analogrelay:
:up: :date: I tested and it does seem like Windows and MSBuild are just fine with using `/` as a directory separator. So we don't need the gross `$(DirSep)` thing. We do need to be diligent and make sure we're using `/` everywhere for paths...

src/Microsoft.AspNetCore.Build/GatherProjectMetadata.cs
```diff
@@ -0,0 +1,62 @@
...
+        [Required]
+        public ITaskItem[] Projects { get; set; }
+
+        [Output]
+        public ITaskItem[] UpdatedProjects { get; set; }
+
+        public override bool Execute()
+        {
+            UpdatedProjects = new ITaskItem[Projects.Length];
+            for (int i = 0; i < Projects.Length; i++)
```
pranavkm: `var`

src/Microsoft.AspNetCore.Build/NuGetResilientPush.cs
```diff
@@ -0,0 +1,130 @@
...
+using NuGet.Commands;
+using NuGet.Common;
+using NuGet.Configuration;
+using MSBuildTask = Microsoft.Build.Utilities.Task;
+
+namespace Microsoft.AspNetCore.Build
+{
+    public class NuGetResilientPush : MSBuildTask
+    {
+        private volatile bool _success = true;
```
pranavkm: Why `volatile`?

src/Microsoft.AspNetCore.Build/ResolvePackagedTool.cs
```diff
@@ -0,0 +1,44 @@
...
+        [Output]
+        public string ToolVersion { get; set; }
+
+        public override bool Execute()
+        {
+            // Find candidates
+            var packageIdDir = Path.Combine(PackagesDir, PackageId);
+            var version = Directory.GetDirectories(packageIdDir)
+                .Select(d => Path.GetFileName(d))
+                .Select(f => NuGetVersion.Parse(f))
```
pranavkm: `TryParse` + `Where(version != null)`?

src/Microsoft.AspNetCore.Build/NuGetResilientPush.cs
```diff
@@ -0,0 +1,130 @@
...
+                if (!localResult)
+                {
+                    _success = false;
+                }
+
+                // Asynchronously play-back the log, but on an un-awaited task so it doesn't block further work
+                logTasks.Add(Task.Run(() =>
+                {
+                    // Lock so that each individual push gets written as a single chunk of log
+                    lock(padlock)
```
pranavkm: `Ctrl k + d`

src/Microsoft.AspNetCore.Build/project.json
```diff
@@ -1,26 +1,31 @@
...
-  "dependencies": {
-    "Microsoft.NETCore.App": {
-      "version": "1.0.0-*",
-      "type": "platform"
+    "dependencies": {
+        "Microsoft.NETCore.App": {
+            "version": "1.0.0-*",
+            "type": "platform"
+        },
+        "NuGet.Versioning": "3.5.0-beta-final",
```
pranavkm: `3.5.0-beta2-1484` - it's part of the RTM build.

src/Microsoft.AspNetCore.Build/project.json
```diff
@@ -1,26 +1,31 @@
...
+        "NuGet.Packaging": "3.5.0-beta-final",
+        "NuGet.Commands": "3.5.0-beta-final",
+        "Newtonsoft.Json": "8.0.3",
```
pranavkm: `9.0.1`

src/Microsoft.AspNetCore.Build/project.json
```diff
@@ -1,26 +1,31 @@
...
     "include": [ "Microsoft.AspNetCore.Build.tasks" ]
   },
 
-  "dependencies": {
-    "Microsoft.NETCore.App": {
-      "version": "1.0.0-*",
-      "type": "platform"
+    "dependencies": {
+        "Microsoft.NETCore.App": {
+            "version": "1.0.0-*",
```
pranavkm: We could pin this to RTM too.

src/Microsoft.AspNetCore.Build/targets/Compile.targets
```diff
@@ -0,0 +1,21 @@
...
+<Project ToolsVersion="14.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+    <!-- Put targets attached to the 'Compile' phase of the Standard Lifecycle here. -->
+
+    <Target Name="DotNetBuild" BeforeTargets="Compile">
+        <ItemGroup>
+            <_ToBuild Include="@(Projects)" Condition="'$(BuildSourceOnly)' != 'true' Or '%(Projects.ProjectGroup)' == 'src'" />
+        </ItemGroup>
+
+        <Delete Files="%(_ToBuild.GeneratedBuildInfoFile)" Condition="Exists(%(_ToBuild.GeneratedBuildInfoFile))" />
+        <WriteLinesToFile
```
pranavkm, line 11: Does it always generate this? We do this conditionally in old-build to avoid depending on git being in the path (plus the cost on dev boxes where it's pretty unnecessary).

src/Microsoft.AspNetCore.Build/targets/Initialize.targets
```diff
@@ -0,0 +1,127 @@
...
+        </ReadLinesFromFile>
+        <Delete Files="$(_TempFile)" />
+    </Target>
+
+    <Target Name="DetectOSPlatform">
+        <DetectOSPlatform>
+            <Output TaskParameter="PlatformName" PropertyName="BuildPlatformName" />
+        </DetectOSPlatform>
+        <PropertyGroup>
+            <NuGetPath Condition="'$(BuildPlatformName)' == 'Windows'">$(MSBuildBinPath)\CoreRun.exe $(MSBuildBinPath)\NuGet.CommandLine.XPlat.dll</NuGetPath>
```
pranavkm: Would 
`<NuGetPath>$(MSBuildBinPath)/corerun $(MSBuildBinPath)/NuGet.CommandLine.XPlat.dll</NuGetPath>`

work all the time?

src/Microsoft.AspNetCore.Build/DetectOSPlatform.cs
```diff
@@ -0,0 +1,36 @@
...
+            {
+                PlatformName = "Linux";
+            }
+            else if (RuntimeInformation.IsOSPlatform(OSPlatform.OSX))
+            {
+                PlatformName = "macOS";
+            }
+            else
+            {
+                PlatformName = "Unknown";
```
cesarblum: Why not return false when the platform can't be determined?

pranavkm:
LKGTM. :watch: for a few of the changes.

src/Microsoft.AspNetCore.Build/DumpMetadata.cs
```diff
@@ -0,0 +1,28 @@
...
+namespace Microsoft.AspNetCore.Build
+{
+    public class DumpMetadata : Task
+    {
+        [Required]
+        public ITaskItem[] Items { get; set; }
+
+        public override bool Execute()
+        {
+            foreach(var item in Items)
```
cesarblum: nit: space after `foreach`

src/Microsoft.AspNetCore.Build/DumpMetadata.cs
```diff
@@ -0,0 +1,28 @@
...
+            {
+                Log.LogMessage(item.ItemSpec);
+                foreach(var metadataName in item.MetadataNames.Cast<string>())
```
cesarblum: nit: space after `foreach`

src/Microsoft.AspNetCore.Build/DumpMetadata.cs
```diff
@@ -0,0 +1,28 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.Build.Framework;
+using Microsoft.Build.Utilities;
+using System.Linq;
```
cesarblum: `System.*` at the top

src/Microsoft.AspNetCore.Build/NuGetInstall.cs
```diff
@@ -0,0 +1,71 @@
...
+            var success = true;
+            var logger = new MSBuildNuGetLogger(Log);
+            foreach (var item in Packages)
+            {
+                // Load the package
+                var path = item.GetMetadata("FullPath");
+                if (!File.Exists(path))
+                {
+                    Log.LogError($"Package file not found: {path}");
+                    success = false;
```
cesarblum: Do you still care about performing remaining work if you hit this condition? Can't you just break here?

src/Microsoft.AspNetCore.Build/NuGetResilientPush.cs
cesarblum: For diagnostic purposes I think you should look at the actual exception and log it's message, even if you're retrying. What if there are different exceptions between failures?

src/Microsoft.AspNetCore.Build/ReplayNuGetLogger.cs
```diff
@@ -0,0 +1,64 @@
...
+        }
+
+        public void LogWarning(string data)
+        {
+            _recordings.Add(l => l.LogWarning(data));
+        }
+
+        public void Replay(ILogger logger)
+        {
+            foreach(var recording in _recordings)
```
cesarblum: nit: space after `foreach`

src/Microsoft.AspNetCore.Build/scripts/KoreBuild.ps1
```diff
@@ -117,8 +117,25 @@ function EnsureMSBuild() {
     }
 }
 
+function EnsureTools() {
```
victorhurdugaci, line 120: Shouldn't this be in the .sh file too?

src/Microsoft.AspNetCore.Build/targets/Initialize.targets
```diff
@@ -0,0 +1,127 @@
...
+            <NuGetPath Condition="'$(BuildPlatformName)' != 'Windows'">$(MSBuildBinPath)\corerun $(MSBuildBinPath)\NuGet.CommandLine.XPlat.dll</NuGetPath>
```
victorhurdugaci: This might work on windows too because I think Windows searches for *.exe if it cannot find the exact name

src/Microsoft.AspNetCore.Build/NuGetResilientPush.cs
analogrelay: It's set to `false` inside the `Parallel.ForEach` which may be run in multiple threads.

src/Microsoft.AspNetCore.Build/project.json
analogrelay: Yeah, it's only there because we have to make this a console app because we need to "publish" it due to limitations in MSBuild Core, but I agree it'll be better to lock it.

src/Microsoft.AspNetCore.Build/targets/Initialize.targets
analogrelay: Possibly. I'll try

src/Microsoft.AspNetCore.Build/targets/Compile.targets
analogrelay, line 11: It does always do this for now. We can re-examine if we want, but I don't think the impact is high enough to make it conditional right now. I can certainly make it conditional on CommitHash having a value (and then we can move that decision to the code that sets CommitHash)

analogrelay, line 11: Writing the file took 15ms, and collecting it took 400ms. I think we should try to embed it every single time we make a build (it's extremely useful) and those are pretty minimal parts of our process (especially since they can be run in parallel when I, eventually, enable parallel build)

src/Microsoft.AspNetCore.Build/targets/Initialize.targets
analogrelay: Indeed it does.

src/Microsoft.AspNetCore.Build/scripts/KoreBuild.ps1
analogrelay, line 120: No, the tool in question is NuGetPackageVerifier which is windows only atm.

src/Microsoft.AspNetCore.Build/targets/Initialize.targets
analogrelay: See @pranavkm 's comment above. Fixed as suggested :)

src/Microsoft.AspNetCore.Build/NuGetResilientPush.cs
pranavkm: We don't do anything besides setting it inside `Parallel.ForEach`. Should be safe to remove the keyword.

analogrelay:
:up: :date: I wrote a quick custom task to handle reading Git commit info. It took that process from 400ms to 58ms, so I think it was worth it :). It also removes our dependency on git.exe and makes us a little more resilient when builds take place outside of a Git repo.

Also moved the tasks into a `Tasks` namespace so we could tell them apart from non-Task classes (since MSBuild doesn't use a common suffix like `Task` :))

pranavkm:
:shipit: . Let's get this in (maybe once the CI is up?) and we can start addressing things as they come up.

analogrelay:
No need to wait for the CI really, nothing builds using this.

PR #98, opened 2016-07-01
Fixing folders and channel for non-Windows

moozzyk:


analogrelay:
:shipit: I noticed the channel thing this afternoon too :). Thanks!

PR #99, opened 2016-07-01
Moving common settings to a separate file

moozzyk:
This makes it possible for custom builds to use the common settings

### DIFF
--- a/src/Microsoft.AspNetCore.Build/targets/makefile.proj
+++ b/src/Microsoft.AspNetCore.Build/targets/makefile.proj
@@ -1,13 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="14.0" InitialTargets="Initialize" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
-    <PropertyGroup>
-        <!-- Product-specific properties -->
-        <DefaultAuthors>Microsoft Open Technologies, Inc.</DefaultAuthors>
-        <ProductStartDate>2015-01-01</ProductStartDate>
-        <BuildQuality>alpha</BuildQuality>
-        <BuildVersion>1.0.1</BuildVersion>
-    </PropertyGroup>
-
+    <Import Project="$(MSBuildThisFileDirectory)\ProductInfo.props" />
     <Import Project="$(MSBuildThisFileDirectory)\StandardLifecycle.targets" />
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNetCore.Build/targets/makefile.proj
```diff
@@ -1,13 +1,6 @@
...
 <Project ToolsVersion="14.0" InitialTargets="Initialize" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
-    <PropertyGroup>
-        <!-- Product-specific properties -->
-        <DefaultAuthors>Microsoft Open Technologies, Inc.</DefaultAuthors>
-        <ProductStartDate>2015-01-01</ProductStartDate>
-        <BuildQuality>alpha</BuildQuality>
-        <BuildVersion>1.0.1</BuildVersion>
-    </PropertyGroup>
-
+    <Import Project="$(MSBuildThisFileDirectory)\Common.settings" />
```
moozzyk: I tested and it and it worked but should we actually use forward slashes in paths?

analogrelay: We can (and possibly should?) use backslashes. I was unsure at first too, but MSBuild Core has code to handle backslashes: https://github.com/Microsoft/msbuild/blob/xplat/src/Shared/FileUtilities.cs#L457

Now I just need to go through and be consistent everywhere :P

analogrelay: Also, can you rename the file to `Common.props`? Generally there are `.targets` and `.props` files in MSBuild, so I'd like to be consistent. I have occasionally seen `.settings`, but I think `.props` is more common :).

Or maybe `ProductInfo.props` to be clear about specifically what is in the file?

analogrelay:
:shipit: after resolving comments.

src/Microsoft.AspNetCore.Build/targets/makefile.proj
moozzyk: Interestingly, I called this file Common.props at the beginning...  I will rename it to ProductInfo.props for now.

moozzyk: while you are at this (consistency) you can also format all files to have 2-space indent :trollface:

analogrelay: I've already begun :). Started noticing now that I'm editing them all in VS

PR #100, opened 2016-07-05
MorkBuild: minor updates and fixes

analogrelay:
- Whitespace cleanup (XML indent -> 2 to match VS defaults)
- Namespace resync (NuGetReplayLogger got accidentally dragged into the Tasks namespace, but it isn't a task)
- Fix hidden messages on Unix (due to using ANSI code for Black, which yields black-on-black in most terminals ;))
- Copy MSBuild/KoreBuild logs to artifacts after build (can't do it in the build because it locks `artifacts` which means we can't really clean it)
- Update `artifacts` final layout to match KoreBuild
- Apparently @victorhurdugaci says we're calling this new MSBuild-based KoreBuild "MorkBuild" and well... 

**ALSO**: Removed KOREBUILD_COMPATIBILITY. Turns out Universe will need to call different targets anyway, which means it needs to "detect" KoreBuild2 already :(. May as well just remove the compat stuff now then.

/cc @pranavkm @victorhurdugaci

### DIFF
--- a/src/Microsoft.AspNetCore.Build/scripts/KoreBuild.ps1
+++ b/src/Microsoft.AspNetCore.Build/scripts/KoreBuild.ps1
@@ -1,28 +1,9 @@
 Write-Host -ForegroundColor Green "Starting KoreBuild 2.0 ..."
 
-if($env:KOREBUILD_COMPATIBILITY -eq "1") {
-    # Rewrite arguments to handle compatibility with 1.0
-    $new_args = $args | ForEach-Object {
-        if($_ -eq "--quiet") {
-            "/v:m"
-        }
-        elseif($_.StartsWith("--")) {
-            # Other unknown switch
-            Write-Warning "Unknown KoreBuild 1.0 switch: $_. If this switch took an argument, you'll have a bad problem :)"
-        } else {
-            $target = [char]::ToUpper($_[0]) + $_.Substring(1)
-            "/t:$target"
-        }
-    }
-    Write-Host -ForegroundColor DarkGray "KoreBuild 1.0 Compatibility Mode Enabled"
-    Write-Host -ForegroundColor DarkGray "KoreBuild 1.0 Command Line: $args"
-    $args = $new_args
-    Write-Host -ForegroundColor DarkGray "KoreBuild 2.0 Command Line: $args"
-}
-
 $RepositoryRoot = Convert-Path (Get-Location)
 $BuildRoot = Join-Path $RepositoryRoot ".build"
 $KoreBuildRoot = (Split-Path -Parent (Split-Path -Parent (Split-Path -Parent $PSScriptRoot)))
+$ArtifactsDir = Join-Path $RepositoryRoot "artifacts"
 
 $env:REPO_FOLDER = $RepositoryRoot
 $env:KOREBUILD_FOLDER = $KoreBuildRoot
@@ -40,13 +21,25 @@ if(Test-Path $KoreBuildLog) {
     del $KoreBuildLog
 }
 
+$MSBuildLog = Join-Path $BuildRoot "korebuild.msbuild.log"
+if(Test-Path $MSBuildLog) {
+    del $MSBuildLog
+}
+
+$MSBuildResponseFile = Join-Path $BuildRoot "korebuild.msbuild.rsp"
+if(Test-Path $MSBuildResponseFile) {
+    del $MSBuildResponseFile
+}
+
+
 function exec($cmd) {
     $cmdName = [IO.Path]::GetFileName($cmd)
     Write-Host -ForegroundColor DarkGray "> $cmdName $args"
     "`r`n>>>>> $cmd $args <<<<<`r`n" >> $KoreBuildLog
-    & $cmd @args >> $KoreBuildLog
-    if($LASTEXITCODE -ne 0) {
-        throw "Command returned exit code $($LASTEXITCODE): '$cmd $args'"
+    & $cmd @args 2>&1 >> $KoreBuildLog
+    $exitCode = $LASTEXITCODE
+    if($exitCode -ne 0) {
+        throw "'$cmdName $args' failed with exit code: $exitCode. See '$ArtifactsDir\korebuild.log' for details"
     }
 }
 
@@ -107,10 +100,13 @@ function EnsureMSBuild() {
             exec dotnet restore "$KoreBuildRoot\src\Microsoft.AspNetCore.Build" -v Detailed
             exec dotnet publish "$KoreBuildRoot\src\Microsoft.AspNetCore.Build" -o "$MSBuildDir\bin\pub" -f "netcoreapp1.0"
         } catch {
+            $ex = $error[0]
             # Clean up to ensure we aren't half-initialized
             if(Test-Path $MSBuildDir) {
                 del -rec -for $MSBuildDir
             }
+
+            throw $ex
         }
     } else {
         Write-Host -ForegroundColor DarkGray "MSBuild already initialized, use -Reset to refresh it"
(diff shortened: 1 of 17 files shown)


src/Microsoft.AspNetCore.Build/scripts/KoreBuild.sh
```diff
@@ -125,6 +124,18 @@ ensure_msbuild
 
 KOREBUILD_TARGETS_ROOT="$KOREBUILD_ROOT/src/Microsoft.AspNetCore.Build/targets"
 
+MSBUILD_RSP=$(mktemp -t koremsbuild)
+
+cat > $MSBUILD_RSP <<ENDMSBUILDARGS
```
pranavkm, line 165: Do you need rsp files in xplat? I thought they were only needed on Windows because cmd had odd argument size restrictions.

src/Microsoft.AspNetCore.Build/targets/Package.targets
```diff
@@ -1,18 +1,15 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="14.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
-    <!-- Put targets attached to the 'Clean' phase of the Standard Lifecycle here. -->
+  <!-- Put targets attached to the 'Clean' phase of the Standard Lifecycle here. -->
 
-    <Target Name="CleanPreviousPackOutput">
```
pranavkm, line 5: Why did this get removed?

src/Microsoft.AspNetCore.Build/targets/ProductInfo.props
```diff
@@ -1,10 +1,10 @@
...
-    <PropertyGroup>
-        <!-- Product-specific properties -->
-        <DefaultAuthors>Microsoft Open Technologies, Inc.</DefaultAuthors>
-        <ProductStartDate>2015-01-01</ProductStartDate>
-        <BuildQuality>alpha</BuildQuality>
-        <BuildVersion>1.0.1</BuildVersion>
-    </PropertyGroup>
+  <PropertyGroup>
+    <!-- Product-specific properties -->
+    <DefaultAuthors>Microsoft Open Technologies, Inc.</DefaultAuthors>
```
pranavkm: `<DefaultAuthors>Microsoft</DefaultAuthors>`

src/Microsoft.AspNetCore.Build/targets/Test.targets
```diff
@@ -1,16 +1,16 @@
...
-    </Target>
+  <Target Name="DotNetTest" BeforeTargets="Test">
+    <ItemGroup>
+      <_ToTest Include="@(TestProjects)" Condition="'$(BuildPlatformName)' == 'Windows' Or '%(TFM_netcoreapp1_0)' == 'true'" />
+    </ItemGroup>
+    <PropertyGroup>
+      <DotNetTestFrameworkArg Condition="'$(BuildPlatformName)' != 'Windows'">-f netcoreapp1.0</DotNetTestFrameworkArg>
+    </PropertyGroup>
+    <Exec
+        Command="dotnet test --configuration &quot;$(Configuration)&quot; --no-build &quot;%(_ToTest.FullPath)&quot; $(DotNetTestFrameworkArg)"
```
pranavkm: 2 spaces.

pranavkm: Similarly in all the other `Exec` things.

pranavkm:
:shipit:

src/Microsoft.AspNetCore.Build/scripts/KoreBuild.sh
analogrelay, line 165: Not technically, but it's a little easier to keep track of arguments if they are the same format in both, IMO.

src/Microsoft.AspNetCore.Build/scripts/KoreBuild.ps1
```diff
@@ -150,6 +150,33 @@ if(Test-Path $MSBuildLog) {
...
 }
 
+$MSBuildArguments = @"
+/nologo
+"$Proj"
+/p:KoreBuildToolsPackages="$ToolsDir"
+/p:KoreBuildTargetsPath="$KoreBuildTargetsRoot"
+/p:KoreBuildTasksPath="$MSBuildDir\bin\pub"
+/fl
+/flp:LogFile="$MSBuildLog";Verbosity=diagnostic;Encoding=UTF-8
```
victorhurdugaci, line 152: Verbosity should be configurable and I don't think it should be diagnostic by default

src/Microsoft.AspNetCore.Build/targets/ProductInfo.props
analogrelay: KoreBuild 1.0 had "Microsoft Open Technologies, Inc." is that incorrect? @Eilon ?

src/Microsoft.AspNetCore.Build/scripts/KoreBuild.sh
```diff
@@ -135,4 +146,9 @@ MSBUILD_LOG="$BUILD_ROOT/korebuild.msbuild.log"
 
 echo -e "${GREEN}Starting build...${RESET}"
 echo -e "${CYAN}> msbuild $PROJ $@${RESET}"
-"$MSBUILD_DIR/bin/pub/corerun" "$MSBUILD_DIR/bin/pub/MSBuild.exe" -nologo $PROJ -p:KoreBuildToolsPackages="$BUILD_DIR" -p:KoreBuildTargetsPath="$KOREBUILD_TARGETS_ROOT" -p:KoreBuildTasksPath="$MSBUILD_DIR/bin/pub/" -fl -flp:logFile="$MSBUILD_LOG;verbosity=diagnostic" "${MSBUILD_ARGS[@]}"
+"$MSBUILD_DIR/bin/pub/corerun" "$MSBUILD_DIR/bin/pub/MSBuild.exe" "${MSBUILD_ARGS[@]}"
```
victorhurdugaci: Delete the rsp file here too? I see that you're doing it in ps1

src/Microsoft.AspNetCore.Build/targets/Package.targets
analogrelay, line 5: Immediately: It was running after compile, which was putting stuff in that directory. I'll look at where to move it. Probably just to "Prepare" (the "Clean" targets are for cleaning the entire repo, which would destroy incremental builds, but cleaning `artifacts` in "Prepare" wouldn't do that because we still build in-place, i.e. in  `bin`/`obj`)

src/Microsoft.AspNetCore.Build/scripts/KoreBuild.ps1
moozzyk, line 152: I think it's fine for the file logger actually - you don't need to run build twice...

src/Microsoft.AspNetCore.Build/targets/ProductInfo.props
pranavkm: That got fixed recently - https://github.com/aspnet/KoreBuild/commit/c9101e20a6c158f38fe8cd32d3024505519c06a1#diff-35569808bff7288226d124655aa58207R30

src/Microsoft.AspNetCore.Build/targets/Package.targets
analogrelay, line 5: It actually already moved to Prepare :) - https://github.com/aspnet/KoreBuild/pull/100/files#diff-50e060fd4d191717e31b8da8bdfbd924R43

(PackagesDir is dead, BuildDir is where things go)

src/Microsoft.AspNetCore.Build/scripts/KoreBuild.ps1
analogrelay, line 152: Yeah, console verbosity is configurable by default using `.\build /v:<verbosity>`. This is for the "ack, everything went wrong and I don't know why" log file which is dumped to `artifacts\korebuild.msbuild.log` :). It's intentionally mega-verbose so that when a failure occurs, you don't have to re-run at a higher verbosity level to see the results. Also we can publish it as an artifact from TeamCity to see the full log from every build as needed.

src/Microsoft.AspNetCore.Build/scripts/KoreBuild.sh
analogrelay: Well... It would also help if we actually USED the RSP here ;P. Will fix both problems.

analogrelay: Also, I'm going to change it to put the RSP in `artifacts` (like the log) so it can be preserved and reviewed if you need to figure out why a build is failing. It can be extremely useful to see the RSP.

analogrelay:
:up: :date:
- Address PR Feedback
- Move all args to "RSP" file and put it in artifacts after the build
- Improve handling of Log/RSP files when builds fail (i.e. make sure they always end up in artifacts, regardless of the build outcome)
- Add `#Requires -Version 4` to PowerShell files just to be cautious

analogrelay:
One more :up: :date: - I added a target that UniverseCoherence can call so it can be isolated from the specific layout of targets.

src/Microsoft.AspNetCore.Build/scripts/KoreBuild.ps1
```diff
@@ -1,28 +1,9 @@
 Write-Host -ForegroundColor Green "Starting KoreBuild 2.0 ..."
 
-if($env:KOREBUILD_COMPATIBILITY -eq "1") {
```
pranavkm, line 3: Why remove this? Isn't this going to be problematic during conversion?

analogrelay, line 3: Turns out there are enough things I want Universe to do differently for KoreBuild 2.0 repos that it's easier to do it on that end.

Issue #30, opened 2016-03-09
Set min required PowerShell version to 4


Eilon:
To avoid people with Windows 7 and PowerShell 3 trying to run the build scripts.

mikes-gh:
I was the one with the above issue.  It was a proxy server issue in the end.  I upgraded to PS 5 by the time I realised the problem but I suspect PS 3 would still work fine.
I think dotnet restore could come back with a better error if nugget sources aren't available.

PR #101, opened 2016-07-05
Require Powershell 4

victorhurdugaci:
Fixes https://github.com/aspnet/KoreBuild/issues/30

cc @anurse

### DIFF
--- a/build/KoreBuild.ps1
+++ b/build/KoreBuild.ps1
@@ -1,3 +1,5 @@
+#requires -version 4
+
 $repoFolder = $env:REPO_FOLDER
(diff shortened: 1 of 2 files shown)


victorhurdugaci:
@pranavkm

pranavkm:
:shipit: 

Would there be a reason not to require 5.0? My dev box (Win 10) has it by default so it removes the risk of accidentally using PS5 features in our scripts that don't work with ps4.0.

victorhurdugaci:
I don't think so. Most systems have 4 installed

PR #104, opened 2016-07-13
Pin the version of NuGet.exe to a version that supports serviceable attribute

pranavkm:


pranavkm:
cc @victorhurdugaci \ @muratg

pranavkm:
This affects dev branches too. Running `build --for-ci verify` fails for repos which repack (such as Common and RazorTooling) since the Latest build of nuget.exe doesn't support `serviceable`. Our CI uses a custom build of nuget.exe which is why it does not see this.

PR #109, opened 2016-08-17
Use ** rather than * for globbing

geirsagberg:
I ran into an issue when building https://github.com/aspnet/JavaScriptServices where the samples were in a lower subdirectory.  Since the glob `samples/*/project.json` would only match projects directly below `samples`, the projects were not being built. Using `**` instead matches all `project.json` in all subdirectories.


BrennanConroy:
Pretty sure we did this on purpose, it avoids building things like /samples/HelloSample/bin/Debug/net451/project.json

davidfowl:
We shouldn't change our default but having it be configurable is ok

geirsagberg:
Gotcha. I'll close the PR for now and re-open it if I can figure out how to make it configurable.

PR #110, opened 2016-08-25
Add ci-test target which allows us to run tests without deleting the src directory

pranavkm:
Makes it a  bit easier to verify how the CI runs these tests + we can get rid of KOREBUILD_NOSRC_EXCLUDE

### DIFF
--- a/build/shade/_k-standard-goals.shade
+++ b/build/shade/_k-standard-goals.shade
@@ -234,6 +234,26 @@ default SAMPLES_PROJECT_GLOB = "samples/*/project.json"
     nuget-resilient-publish sourcePackagesDir='${BUILD_DIR}' nugetFeed='${E("NUGET_PUBLISH_FEED")}' if='!string.IsNullOrEmpty(E("NUGET_PUBLISH_FEED"))'
   -}
 
+#ci-test
+  @{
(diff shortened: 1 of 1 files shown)


pranavkm:
cc @anurse \ @victorhurdugaci

pranavkm:
Follow up: https://github.com/aspnet/Universe/pull/455

victorhurdugaci:
Can you check that this actually works with dotnet watch? That repo expects the src folder to be built. Also, simply removing the the "src" folder from global.json will not prevent sources from being built

pranavkm:
I had to tweak dotnet-watch to not use DOTNET_BUILD_VERSION and create the artifacts directory if it didn't already exist: https://github.com/aspnet/dotnet-watch/pull/153. That gets this target building with this repo and will start running tests against packages compiled previously (as opposed to what it does not). 

We don't compile src as part of this target, so that's not an issue.

victorhurdugaci:
Does that break local development? Will local tests run against what you compile or the published packages?

pranavkm:
What you compile as long as you have an artifacts directory. If you specifically run ci-test, it would run it against published packages.

victorhurdugaci:
cool!

PR #116, opened 2016-09-27
Allow configuring the shared runtime via an environment variable

pranavkm:
cc @anurse \ @BrennanConroy

### DIFF
--- a/build/KoreBuild.sh
+++ b/build/KoreBuild.sh
@@ -40,6 +40,8 @@ version=$(<$versionFile)
 
 [ -z "$KOREBUILD_DOTNET_CHANNEL" ] && KOREBUILD_DOTNET_CHANNEL=rel-1.0.0
 [ -z "$KOREBUILD_DOTNET_VERSION" ] && KOREBUILD_DOTNET_VERSION=$version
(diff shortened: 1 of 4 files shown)


build/KoreBuild.sh
```diff
@@ -40,6 +40,8 @@ version=$(<$versionFile)
 
 [ -z "$KOREBUILD_DOTNET_CHANNEL" ] && KOREBUILD_DOTNET_CHANNEL=rel-1.0.0
 [ -z "$KOREBUILD_DOTNET_VERSION" ] && KOREBUILD_DOTNET_VERSION=$version
+[ -z "$KOREBUILD_DOTNET_SHARED_RUNTIME_VERSION " ] && KOREBUILD_DOTNET_VERSION=1.0.0
```
BrennanConroy: These need to be the same variable...

BrennanConroy: Or wait, we don't have a shared runtime install here?

pranavkm: Fixed

BrennanConroy:
:shipit:

PR #117, opened 2016-09-28
Run package verification rules when possible

pranavkm:
cc @ryanbrandenburg


ryanbrandenburg (approved):
:shipit: but I'm curious why it can't be run on Linux?

pranavkm:
The verifier is desktop only. FIled https://github.com/aspnet/BuildTools/issues/103 to track the fix for it but we can get something going until that's sorted.

PR #118, opened 2016-09-28
Move dotnet-pack calls into separate target

natemcmaster:
Make it easier to override the whole default packing mechanism. aspnet/DotnetTools and aspnet/EntityFramework.Tools actually have to remove packages from artifacts/build that we don't actually want in build output.

cc @pranavkm


pranavkm:
:shipit:

PR #122, opened 2016-10-04
Update shared runtime to 1.1.0-preview1-001090-00

pranavkm:


pranavkm:
FYI @muratg 

We probably need a way for the gate to grab the latest build of the shared runtime and update a file the way it does for the CLI. But for now this will do

muratg:
Cool. Though I expect we may need to do this a couple times more before we release.

PR #123, opened 2016-10-05
Fix build break in most repos on Linux

dougbu:
- strings containing spaces are never empty
- incorrectly left `$KOREBUILD_DOTNET_SHARED_RUNTIME_VERSION` undefined
- later `-d` check confirmed only that the parent directory existed
- script skipped the `--share-runtime` install

nit: use same channel for regular install as `KoreBuild.ps1`


dougbu:
FYI only. I've tested this locally and this **cannot** make Travis worse.

dougbu:
6c026ce

PR #126, opened 2016-10-17
Run NPM installs in series, not parallel

SteveSandersonMS:
I'm seeing intermittent build failures on the JavaScriptServices repo when it builds on Windows. Since there have been many reports of NPM instability when run in parallel on Windows, this change makes NPM installs run in series. Obviously it will take longer, but hopefully it will produce more consistent results. We won't know whether this actually fixes the issue until this runs in CI.


pranavkm:
:shipit:

PR #128, opened 2016-10-31
Adding support for --shared-runtime to dotnet-install.sh

pranavkm:
The 1.0.0-preview2.1 bash script doesn't support --shared-runtime required to install
Microsoft.NETCore.App 1.0.0. Back porting from the 1.0.0 dotnet-install script which is
incompatible with the older runtime's download structure.

### DIFF
--- a/build/dotnet/dotnet-install.sh
+++ b/build/dotnet/dotnet-install.sh
@@ -377,8 +377,12 @@ construct_download_link() {
     local specific_version=${4//[$'\t\r\n']}
     
     local osname=$(get_current_os_name)
-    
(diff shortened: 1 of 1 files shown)


pranavkm:
cc @NTaylorMullen \ @BrennanConroy

BrennanConroy:
So next time we update this file how are we going to avoid back porting --shared-runtime again?

pranavkm:
They have this fixed in the 1.0.0 CLI scripts. But that expects a different structure than the one they used for the preview2.1 scripts. So hopefully we wouldn't need to backport things in the future.

NTaylorMullen:
:shipit:

pranavkm:
@BrennanConroy ping

BrennanConroy:
:shipit:

PR #136, opened 2016-11-17
Install NET451 reference assemblies

pranavkm:


pranavkm:
cc @natemcmaster \ @JunTaoLuo. Should fix an issue we're seeing building Kestrel in rel/1.0.3

natemcmaster:
:shipit:  - do we need this on dev too?

pranavkm:
We already do. We weren't in 1.0.0 KoreBuild which is where `rel/1.0.1` started from.

PR #145, opened 2016-12-15
Update common.props to import targets containing workarounds

natemcmaster:
This centralizes workarounds for bugs in Microsoft.NET.Sdk so that we can share them across all builds. We want to minimize workarounds in local repos.

cc @prafullbhosale may require a few tweaks for aspnet/Scaffolding.


template/build/common.props
```diff
@@ -0,0 +1,8 @@
+<Project>
```
prafullbhosale, line 1: Where does this file get imported?

prafullbhosale: NM, this should get imported by individual projects in the repository.

prafullbhosale (approved):

PR #146, opened 2016-12-15
Always install the SDK specified in KoreBuild

pranavkm:
We will rely on the global.json to pick up the right version of the CLI.


pranavkm:
Spoke to @natemcmaster offline and realized we might not need this on an ongoing basis. I'll workaround this for now and we can revive this code if we truly need it.

PR #148, opened 2016-12-15
Add support for injecting properties from Universe into the build

natemcmaster:
Companion change to https://github.com/aspnet/Universe/pull/467


natemcmaster:
Closing. We don't need this for now.

PR #149, opened 2016-12-16
Resolve versions for ProjectReferences and floating package references

natemcmaster:
Unblock our build.

Workaround https://github.com/NuGet/Home/issues/3953 and https://github.com/NuGet/Home/issues/4063

cc @rohit21agrawal @jgoshi @bricelam


natemcmaster:
Merging to unblock CI

Issue #152, opened 2016-12-19
Two issues when running 'build initialize' in Razor sources


Yves57:
I just cloned the Razor sources with Visual Studio 2017 RC (aspnet/Razor#903).
When I tried to compile the Razor solution for the first time in Visual Studio, it requested to run `build initialize` in a command line.

But I had two problems with the script:

- It failed to download the SDK. It seems to be a timeout problem (I have a slow connection) because the URL is valid. And as workaround, I succeed to insert `$HttpClient.Timeout = New-TimeSpan -Hours 1` line 150 of 'dotnet-install.ps1'.
![timeout](https://cloud.githubusercontent.com/assets/11701094/21311797/673f6340-c5e9-11e6-91c4-51b8f92bc802.png)

- It failed in the installation script because of a small mistake in 'common.props'. The script search the index of a '.' (dot) while it is a ',' (comma) on my computer.
![syntax](https://cloud.githubusercontent.com/assets/11701094/21311801/6d03e616-c5e9-11e6-9880-4ed2019255ca.png)

Eilon:
@NTaylorMullen can you take a look?

NTaylorMullen:
Hmm what version of VS2017 are you using @Yves57?

Yves57:
@NTaylorMullen 
Probably the latest (downloaded sunday 18).
From the About dialog:
```
Microsoft Visual Studio Community 2017 RC
Version 15.0.26014.0 D15REL
Microsoft .NET Framework
Version 4.6.01586

Installed Version: Community

Visual Basic 2017 RC   00369-50000-00000-AA466
Microsoft Visual Basic 2017 RC

Visual C# 2017 RC   00369-50000-00000-AA466
Microsoft Visual C# 2017 RC

Application Insights Tools for Visual Studio Package   8.4.01118.2
Application Insights Tools for Visual Studio

ASP.NET and Web Tools 2016   15.0.21206.0
ASP.NET and Web Tools 2016

ASP.NET Web Frameworks and Tools 2013   5.2.41123.0
For additional information, visit https://www.asp.net/

Azure App Service Tools v2.9.6   15.0.21130.0
Azure App Service Tools v2.9.6

Common Azure Tools   1.8
Provides common services for use by Azure Mobile Services and Microsoft Azure Tools.

GitHub.VisualStudio   2.1.1.4
A Visual Studio Extension that brings the GitHub Flow into Visual Studio.

JavaScript Language Service   2.0
[...]

PR #154, opened 2017-01-06
Fix globalization issue with generating build numbers

natemcmaster:
Fixed part of https://github.com/aspnet/KoreBuild/issues/152#issuecomment-269346831

### DIFF
--- a/build/common.props
+++ b/build/common.props
@@ -19,9 +19,12 @@
 
     <!-- version options -->
     <PropertyGroup>
-        <!-- create second-based build number for local builds -->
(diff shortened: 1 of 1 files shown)


natemcmaster:
@Yves57 I tested this in French and it should solve the comma vs period issue you found in the timestamp formats

Yves57:
Test on my computer:
1) Downloaded today the latest Razor repository sources as ZIP
2) Uncompressed
3) Called `build initialize` --> script error
4) Replaced the incorrect lines by your fix in `common.props`
5) Called `build initialize` again --> success!

Thank you :-)

Please note that I had not the timeout problem (second problem) because the SDK was already installed.

NTaylorMullen (commented):
Solid.

PR #156, opened 2017-01-12
Print link to shared runtime installer

natemcmaster:
VS 2017 doesn't, by default, use the same sharedfx that korebuild will install. This change prints a link to the downloader required to make VS test work on 2017 projects.


BrennanConroy:
Doesn't VS2015 have the same issue? Whenever we get a new runtime I have to copy the shared folder from appdata to program files if I want to work in VS

natemcmaster:
I can but this on both dev and feature/msbuild

natemcmaster:
🔔 rebased on dev

BrennanConroy:
:shipit:

PR #157, opened 2017-01-13
Remove usage of -parallel flag on dotnet-test

natemcmaster:
dotnet-test no longer supports -parallel. Any tests that require non-parallel test runs should configure this via code instead.

Example:
```c#
[assembly: CollectionBehavior(DisableTestParallelization = true)]
```
cref https://xunit.github.io/docs/running-tests-in-parallel.html
cref https://xunit.github.io/docs/configuring-with-xml

cc @bricelam (EF Core SQL Server tests)
cc @pranavkm we should be able to remove this from aspnetci config

### DIFF
--- a/build/shade/_dotnet-test.shade
+++ b/build/shade/_dotnet-test.shade
@@ -22,15 +22,7 @@ default framework = ''
 @{
     var projectFolder = Path.GetDirectoryName(projectFile);
     var projectName = Path.GetFileName(projectFolder);
-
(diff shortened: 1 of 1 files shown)


pranavkm (approved):
The flag is pretty unreliable since we have to set it in local builds. The attribute is the way to go.

bricelam:
Filed an issue to react in EF.

natemcmaster:
cc @cesarbs

PR #158, opened 2017-01-30
Use makefile.proj if available

pranavkm:


natemcmaster:
`makefile.proj` will break all shorthand dotnet commands since it matches the `*.*proj` pattern. This would require you to explicitly call `dotnet build Repo.sln` now and `dotnet restore Repo.sln`.

As an alternative, consider adding a msbuild project into KoreBuild that conditionally imports `makefile.targets` from the repo folder if it exists. This has the added benefit of allowing us to start moving the Sake lifecycle into an MSBuild project.

Concept:
in KoreBuild
```xml
<Project>
    <Target Name="Build" />
    <Import Project="$(RepoDir)\makefile.targets" Condition="Exists('$(RepoDir)\makefile.targets')" />
</Project>
```

in repo
```xml
<Project>
   <Target Name="AfterBuild" AfterTargets="Build" />
</Project>
```

pranavkm:
build/makefile.proj if the file in the root affects builds? Presumably that solves and moves more build related files to the build dir.

natemcmaster:
I like the idea of putting it in the build/ folder. How about build/repo.targets ?

pranavkm:
🆙 📅  with corresponding changes to https://github.com/aspnet/CoreCLR/pull/149/

build/targets/standard-lifecycle.targets
```diff
@@ -0,0 +1,17 @@
+<Project>
```
natemcmaster, line 1: Should this be set to `DefaultTargets="Default"`?

build/targets/legacy-lifecycle.targets
```diff
@@ -0,0 +1,158 @@
...
+      <SakeExecutable Condition="'$(OS)'!='Windows_NT'">mono $(SakeExecutable)</SakeExecutable>
+      <MakeFilePath>$(ShadeDirectory)makefile.shade</MakeFilePath>
+      <MakeFilePath Condition="!Exists('$(RepositoryRoot)makefile.shade')">$(RepositoryRoot)makefile.shade</MakeFilePath>
+
+      <ShadeTarget Condition="'$(NoRestore)'=='true'">--no-restore $(ShadeTarget)</ShadeTarget>
+      <ShadeTarget Condition="'$(MakeQuiet)'=='true'">--quiet $(ShadeTarget)</ShadeTarget>
+    </PropertyGroup>
+
+    <Exec 
+      Command="$(SakeExecutable) -I $(KoreBuildDirectory)shade -f $(MakeFilePath) $(ShadeTarget)"
```
natemcmaster, line 155: I did this exact same work a few weeks ago (bu never finished it) and ran into issues with relative vs absolute paths. Sake.exe didn't recognize absolute paths for the `-f` argument. Have you run into this issue here?

build/KoreBuild.ps1
```diff
@@ -91,13 +93,9 @@ if (!(Test-Path "$koreBuildFolder\Sake"))
...
-
-Write-Host "Using makefile: $makeFilePath"
-
 $env:KOREBUILD_FOLDER=$koreBuildFolder
-&"$koreBuildFolder\Sake\0.2.2\tools\Sake.exe" -I $koreBuildFolder\shade -f $makeFilePath @args
+$makeFileProj = "$koreBuildFolder/targets/makefile.proj"
+if ($allparams) {
+    $targets = '/t:' + $allparams.Replace(' ', ';')
+}
+dotnet msbuild $makeFileProj /p:KoreBuildDirectory="$koreBuildFolder\" /p:RepositoryRoot="$repoFolder\" $targets
```
natemcmaster: $targets needs to be surrounded with double quotes so multiple targets doesn't terminate the line.

Suggested adding these parameters:
/maxcpucount /fileLogger /fileloggerparameters:LogFile=artifacts/msbuild.log /detailedsummary

Also, consider making an rsp file (artifacts/msbuild.rsp) and executing "dotnet msbuild @artifacts/msbuild.rsp" instead.

Nit: add newline at end of file.

natemcmaster (changes requested):
👏 I like where this is going. A few changes suggested.

build/KoreBuild.ps1
pranavkm: I actually tried multiple targets and it works fine - primarily because the `$targets` is a variable and not code in the file with semicolons.

build/targets/legacy-lifecycle.targets
pranavkm: `KoreBuildDirectory` is root relative - notice I had to concat the path on Line 145 with `RepositoryRoot` to make it absolute. Btw, this brought up an interesting issue with using back slashes in our code - the things we invoke on xplat don't like back slashes (for instance `mono build\sake.exe`). Calling `Path.GetFullPath` helps, but it'd be much better to standardize on forward slashes in our targets.

build/targets/standard-lifecycle.targets
pranavkm: This gets included by `makefile.proj` which has the DefaultTarget. Would we need one here?

natemcmaster: nope, you're right. I just didn't see that

build/targets/legacy-lifecycle.targets
natemcmaster: gotcha. was just checking :)

build/KoreBuild.ps1
natemcmaster: glad this PowerShell quirk works in our favor.

pranavkm:
🆙 📅

build/KoreBuild.sh
```diff
@@ -110,10 +114,28 @@ if [ ! -f $nugetPath ]; then
...
+/detailedsummary
+"$makeFileProj"
+/p:KoreBuildDirectory="$koreBuildFolder/"
+/p:RepositoryRoot="$repoFolder/"
+/fl
+-flp:LogFile="$msbuildLogFile";Verbosity=diagnostic;Encoding=UTF-8
+$targets
+ENDMSBUILDARGS
+
+dotnet msbuild @"$msbuildResponseFile"
```
natemcmaster, line 141: One more thought. Right now this converts $@ to a list of targets. How would we go about flowing through properties? e.g. `/p:Configuration=Release`.

natemcmaster (approved):

natemcmaster:
:bell: can we get this in soon? I want to rebase #159 off of this work.

pranavkm:
Merged. Let's see how this goes.

pranavkm:
Ok - need to tweak this so that it allows calling targets that aren't declared in KoreBuild (for instance Universe has different targets). I'll play around with this and resend.

pranavkm:
I'll send an update to this later.

PR #159, opened 2017-01-31
Build the sln file instead of individual project files

natemcmaster:
Perf improvement: only builds each project once during the 'build-compile' target.


build/shade/_k-standard-goals.shade
```diff
@@ -151,24 +150,11 @@ default SAMPLES_PROJECT_GLOB = "samples/*/*.csproj"
...
-    {
-        projectGlobs.AddRange(Files.Include(TEST_PROJECT_GLOB).ToList());
-    }
-    if (!BuildSrcOnly && Directory.Exists("samples") && Directory.EnumerateFiles("samples", "*.csproj", SearchOption.AllDirectories).Any())
-    {
-        projectGlobs.AddRange(Files.Include(SAMPLES_PROJECT_GLOB).ToList());
     }
 
-    if (projectGlobs.Any())
+    foreach (var sln in Files.Include(Path.Combine(BASE_DIR, "*.sln")))
```
pranavkm: Could we just build `$repoName.sln` - seems like a convention all our repos follow. We could throw an error if there happen to be more than one sln.

pranavkm (approved):

build/shade/_k-standard-goals.shade
natemcmaster: Yeah, that could be fine. I know off the top of my head that Diagnostics has DiagnosticPages.sln. Are there others?

JunTaoLuo: HttpSysServer still has WebListener.sln though I think WebListener as a repo name will redirect properly

natemcmaster: We also need to change that rename in aspnet/Universe's list of repos.

pranavkm: > We could throw an error if there happen to be more than one sln.

Meant to say we could throw an error to say we couldn't find the sln if we don't find it.

natemcmaster: We could also just invoke "dotnet build" and let the CLI find the .sln file to use. If there are more than one files matching `*.*proj` or `*.sln` it will fail on its own.

pranavkm: | Repo | Solutions |
|------|-----------|
|Antiforgery|Antiforgery.sln|
|AzureIntegration|AzureIntegration.sln|
|BasicMiddleware|BasicMiddleware.sln|
|BrowserLink|BrowserLink.sln|
|Caching|Caching.sln|
|Common|Common.sln|
|Configuration|Configuration.sln|
|CORS|CORS.sln|
|DataProtection|DataProtection.sln|
|DependencyInjection|DependencyInjection.sln|
|Diagnostics|DiagnosticsPages.sln|
|DotNetTools|DotNetTools.sln|
|EntityFramework|EntityFramework.sln|
|EntityFramework.Tools|EntityFramework.Tools.sln|
|Entropy|Entropy.sln|
|EventNotification|EventNotification.sln|
|FileSystem|FileSystem.sln|
|Hosting|Hosting.sln|
|HtmlAbstractions|HtmlAbstractions.sln|
|HttpAbstractions|HttpAbstractions.sln|
|HttpSysServer|WebListener.sln|
|Identity|Identity.sln|
|IISIntegration|IISIntegration.sln|
|JsonPatch|JsonPatch.sln|
|KestrelHttpServer|KestrelHttpServer.sln|
|Localization|Localization.sln|
|Logging|Logging.sln|
|MetaPackages|MetaPackages.sln|
|Microsoft.Data.Sqlite|Microsoft.Data.Sqlite.sln|
|MusicStore|MusicStore.sln|
|Mvc|Mvc.NoFun.sln Mvc.sln|
|MvcPrecompilation|RazorViewCompilation.sln|
|Options|Options.sln|
|PlatformAbstractions|PlatformAbstractions.sln|
|Proxy|Proxy.sln|
|Razor|Razor.sln|
|ResponseCaching|ResponseCaching.sln|
|Routing|Routing.sln|
|Scaffolding|Scaffolding.sln|
|Security|Security.sln|
|ServerTests|ServerTests.sln|
|Session|Session.sln|
|SignalR|SignalR.sln|
|StaticFiles|StaticFiles.sln|
|Testing|Testing.sln|
|WebSockets|WebSockets.sln|

pranavkm: MvcPrecompilation, Diagnostics and HttpSysServer are the only outliers here. Should be easy to fix them.

natemcmaster: The only problematic thing is...how do you determine "$repoName"? We can't use directory name as it's not guaranteed....

pranavkm: That's true. I guess we should get Mvc to move the second sln (the NoFun one) out of the repo root as part of the migration. We'll switch to calling `dotnet build` and have it fail in the non-common cases.

FYI @NTaylorMullen

pranavkm (approved):

build/shade/_k-standard-goals.shade
NTaylorMullen: Bummer, was working around some issues with having multiple sln's in Mvc's [migrated makefile](https://github.com/aspnet/Mvc/blob/dbcb56f09b97d6ac7c05ace96bd7bf6d99331c57/makefile.shade#L8). Could we take the approach of having a parameter that we can set to override the default sln name?. It'd be a shame to hide sln's that we actually use from users who clone our repos.

pranavkm: Instead of a parameter, we could build a well known file (`build.proj`?) if it's present. We can't run `dotnet build` regardless.

natemcmaster: I think this problem can be solved by the work Pranav is doing on https://github.com/aspnet/KoreBuild/pull/158. The default Korebuild build step could include:

```xml
<SolutionsToBuild Include="$(RepoRoot)\*.sln" />
```
MVC and other repos could customize by adding exclusions to the `build/repo.targets` file:

```xml
<SolutionsToBuild Remove="$(RepoRoot)\Mvc.NoFun.sln" />
```

natemcmaster:
Almost there: something in Korebuild is attempting to use obj/*.dll...I think.

> CSC : error CS2012: Cannot open 'C:\dev\Universe\DependencyInjection\src\Microsoft.Extensions.DependencyInjection\obj\Debug\netstandard1.1\Microsoft.Extensions.DependencyInjection.dll' for writing -- 'The process cannot access the file 'C:\dev\Universe\DependencyInjection\src\Microsoft.Extensions.DependencyInjection\obj\Debug\netstandard1.1\Microsoft.Extensions.DependencyInjection.dll' because it is being used by another process.' [C:\dev\Universe\DependencyInjection\src\Microsoft.Extensions.DependencyInjection\Microsoft.Extensions.DependencyInjection.csproj] [C:\dev\Universe\DependencyInjection\.build\targets\makefile.proj]

Doesn't happen when you call "dotnet build" on the solution, w/o korebuild

natemcmaster:
Nvm, it's dotnet/sdk. https://github.com/dotnet/sdk/issues/739

build/targets/legacy-lifecycle.targets
```diff
@@ -1,162 +1,23 @@
...
-      <ShadeTarget Condition="'$(NoRestore)'=='true'">--no-restore $(ShadeTarget)</ShadeTarget>
-      <ShadeTarget Condition="'$(MakeQuiet)'=='true'">--quiet $(ShadeTarget)</ShadeTarget>
+      <SakeTargets Condition="'$(SakeTargets)' != '' ">$(SakeTargets.Replace(':', ' '))</SakeTargets>
+      <SakeTargets Condition="'$(NoRestore)'=='true'">--no-restore $(SakeTargets)</SakeTargets>
+      <SakeTargets Condition="'$(MakeQuiet)'=='true'">--quiet $(SakeTargets)</SakeTargets>
     </PropertyGroup>
 
     <Exec
-      Command="$(SakeExecutable) -I $(KoreBuildDirectory)shade -f $(MakeFilePath) $(ShadeTarget)"
+      Command="$(SakeExecutable) -I &quot;$(MSBuildThisFileDirectory)../shade&quot; -f $(MakeFilePath) $(SakeTargets)"
```
pranavkm: 👏

pranavkm (approved):

build/targets/common.props
```diff
@@ -1,6 +1,7 @@
 <Project>
   <PropertyGroup>
-    <BuildDir>$(RepositoryRoot)\build</BuildDir>
-    <ArtifactsDir>$(BuildDir)\artifacts</ArtifactsDir>
+    <RepositoryRoot Condition="'$(RepositoryRoot)'==''">$(MSBuildThisFileDirectory)..\..\</RepositoryRoot>
```
dougbu: This location can easily be incorrect, especially in (nested) test directories. Why isn't this using `$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), "$(Repository).sln")` or something similar?

build/targets/standard-lifecycle.targets
```diff
@@ -1,17 +1,44 @@
...
-  <Target Name="Default" DependsOnTargets="Build" />
 
-  <Target Name="Build" DependsOnTargets="Clean;Initialize;Compile;Test;Verify;Publish" />
+  <PropertyGroup>
+    <Configuration>Debug</Configuration>
+  </PropertyGroup>
 
-  <Target Name="Clean" />
+  <ItemGroup>
+    <SolutionsToBuild Include="$(RepositoryRoot)\*.sln" />
```
dougbu: This won't work well at least for Mvc. Why not `$(RepositoryRoot)\$(RepoName).sln`?

build/shade/_k-standard-goals.shade
```diff
@@ -32,6 +32,18 @@ default SAMPLES_PROJECT_GLOB = "samples/*/*.csproj"
...
 }
 
+functions
+  @{
+    // convenince method as we incrementally move to MSBuild standard lifecycle
+    void MSBuild(string arguments)
+    {
+      Dotnet("msbuild /nologo"
+             + " \"" + Path.Combine(Directory.GetCurrentDirectory(), ".build/targets/makefile.proj") + "\""
+             + " \"/p:LifecycleType=Standard;Configuration=" + E("Configuration") + "\""
```
dougbu: With changes to `build/targets/makefile.proj`, does this command-line lifecycle choice still work?

dougbu (changes requested):

build/targets/common.props
natemcmaster: MSBuildThisFileDirectory refers to the location of .build/targets/common.props, not the location of test projects. This value isn't affect by nested test directories.

build/targets/standard-lifecycle.targets
natemcmaster: The goal here is to make the default case work, which is that we have one sln per repo. MVC and any other solution that has multiple solutions would put something like this in to build/repo.targets:
```xml
<SolutionsToBuild Remove="Mvc.NoFun.sln" />
```

We don't have a good way to determine "RepoName" :-/ as the folder name containing the source code is not guaranteed to match the sln name.

build/shade/_k-standard-goals.shade
natemcmaster: For now, yes, you can still express `build verify` and `build --for-ci build-compile` etc.

natemcmaster:
:up: :date: Removing the [blocked] label. The race condition in solution builds has been fixed.

natemcmaster:
I added two changes to be aware of.

(1) the tool dependencies (Sake, NuGetPackageVerifier, and NETFrameworkReferenceAssemblies) restore into the regular nuget cache instead of into the .build folder
(2) Update to CLI 1.0.0-rc4-004777

We'll need to update other repos. Any project that has multiple \*.sln files or has projects in the solution file that doen't build with dotnet-CLI will need to be updated. Razor, MVC, and possibly others.

natemcmaster:
cc @pakrym

pakrym:
Wouldn't calling pack on each project try to rebuild everything again?

natemcmaster:
At the moment yes it will build twice. This is an incremental step towards eliminating the Sake lifecycle. Some of our repos have custom packaging steps that we need to port to MSBuild. 

This has the advantage of only rebuilding projects once during the pack phase. MSBuild will batch the builds during pack.

In a follow up PR, we can move even more into MSBuild so each project should only compile once per build.cmd run.

pakrym:
👍 LGTM

PR #160, opened 2017-02-01
Remove TFM parsing from xunit test target

natemcmaster:
Requires update to projects that use .NET Framework in test projects. These should be updated to only include .NET Framework when building/testing on Windows.


natemcmaster:
I made the changes affected repos to ensure net451 tests are only compiled on Windows. Will push once this is merged

natemcmaster:
:x: blocked. Conditional `TargetFrameworks` properties cause VS to explode

image

natemcmaster:
When this https://github.com/dotnet/roslyn-project-system/pull/1444 makes it into a build, we can return to conditioning TargetFrameworks on OS.

natemcmaster:
🔔 Updated and rebased. Ready for review.

build/targets/standard-lifecycle.targets
```diff
@@ -125,7 +159,51 @@ that matches "$(RepositoryRoot)/shared/*.Sources".
...
+    <!-- TODO if VS Test doesn't make minimal output the default log setting, we can set that here. cref https://github.com/Microsoft/vstest/issues/301 -->
+    <VSTestLogger Condition=" '$(VSTestLogger)'=='' AND '$(TEAMCITY_VERSION)' != '' ">trx</VSTestLogger>
+    <IgnoreFailingTestProjects>false</IgnoreFailingTestProjects>
+    <ContinueOnTestError Condition="'$(KOREBUILD_IGNORE_DOTNET_TEST_EXIT_CODE)' == '1'">true</ContinueOnTestError>
+  </PropertyGroup>
+
+  <Target Name="TestProjects">
+
+    <ItemGroup Condition="'@(ProjectsToTest)'==''">
+      <!-- put unit test projects ahead of functional tests -->
```
pranavkm: Do we actually have this as a convention? `Tests` -> Functional? Otherwise this is kinda misleading.

pranavkm: Plus we don't super care about ordering. Ideally we should try and these in parallel.

build/targets/standard-lifecycle.targets
```diff
@@ -125,7 +159,51 @@ that matches "$(RepositoryRoot)/shared/*.Sources".
...
+      <ProjectsToTest Include="$(RepositoryRoot)test\*\*.Tests.csproj" Exclude="$(ExcludeFromTest)" />
+      <ProjectsToTest Include="$(RepositoryRoot)test\*\*.Test.csproj" Exclude="$(ExcludeFromTest)" />
+      <ProjectsToTest Include="$(RepositoryRoot)test\*\*.csproj" Exclude="@(ProjectsToTest);$(ExcludeFromTest)" />
```
pranavkm: Wouldn't this suffice? Also, why not make `ExcludeFromTest` an item group?

build/targets/standard-lifecycle.targets
```diff
@@ -125,7 +159,51 @@ that matches "$(RepositoryRoot)/shared/*.Sources".
...
+
+Runs the VSTest on all projects in the ProjectsToTest itemgroup.
+###################################################################
+-->
+
+  <PropertyGroup>
+    <!-- TODO if VS Test doesn't make minimal output the default log setting, we can set that here. cref https://github.com/Microsoft/vstest/issues/301 -->
+    <VSTestLogger Condition=" '$(VSTestLogger)'=='' AND '$(TEAMCITY_VERSION)' != '' ">trx</VSTestLogger>
+    <IgnoreFailingTestProjects>false</IgnoreFailingTestProjects>
+    <ContinueOnTestError Condition="'$(KOREBUILD_IGNORE_DOTNET_TEST_EXIT_CODE)' == '1'">true</ContinueOnTestError>
```
pranavkm: Should this be `IgnoreFailingTestProjects`? `ContinueOnTestError` isn't used elsewhere.

pranavkm (approved):

build/targets/standard-lifecycle.targets
natemcmaster: Yup, good catch. Missed that rename.

build/targets/standard-lifecycle.targets
natemcmaster: This is something EF has requested. Running unit tests first to catch errors early is nice, esp when functional tests can take a long time to run.

build/targets/standard-lifecycle.targets
natemcmaster: And item group might be better. I ran into this https://github.com/Microsoft/msbuild/issues/1576 which means setting ExcludeFromTest in repo.targets correctly is not easy.

natemcmaster: Hmm, nvm. ItemGroups and remove syntax suffer from the same issue.

build/targets/standard-lifecycle.targets
pranavkm: Can we explicit about it - requiring functional tests to include the term `FunctionalTests`? Alternatively have `repo.props` determine order if it's necessary

build/targets/DefaultItems.targets
```diff
@@ -0,0 +1,25 @@
...
+    <Solutions Remove="@(ExcludeSolutions)" />
+
+    <ProjectsToPack Include="$(RepositoryRoot)src\*\*.csproj" />
+    <ProjectsToPack Remove="@(ExcludeFromPack)" />
+
+    <SharedSourceDirectories Include="$([System.IO.Directory]::GetDirectories(&quot;$(SharedSourcesFolder)&quot;, '*.Sources'))" Condition="Exists('$(SharedSourcesFolder)')" />
+    <SharedSourceDirectories Remove="@(ExcludeSharedSourceDirectories)" />
+
+    <!-- put unit test projects ahead of functional tests -->
+    <ProjectsToTest Include="$(RepositoryRoot)test\*\*.Tests.csproj" />
```
pranavkm:
```xml
<ItemGroup Condition="'@(ProjectsToTest)'==''">
  <_FunctionalTests Include="$(RepositoryRoot)test\*FunctionalTest*.csproj" />
  <ProjectsToTest Include="$(RepositoryRoot)test\*Test*.csproj" Exclude="@(_FuctionalTests)" />
  <ProjectsToTest Include="@(_FunctionalTests)"  />
 ...
</ItemGroup>
```

natemcmaster: 👍 

Although, the thinking behind not doing `Condition="'@(ProjectsToTest)'==''"` is that the most common scenario here is that we will want to exclude a project from testing `ExcludeFromTest` that was picked up by default globs. Redefining the default globs is cumbersome.

pranavkm: Sure, we could add it if we need it later on.

PR #164, opened 2017-02-07
Do not assume content of the SAMPLES_PROJECT_GLOB variable

dougbu:
- samples folder may exist _only_ to enable use of a modified SAMPLES_PROJECT_GLOB

Other alternatives to the `Directory.Exists("samples")` approach:
- change default glob to something like `sam*ples/*/*.csproj`
  - that wouldn't require the samples folder but may in the future match something unintended
- update PathResolver.shade in Sake to avoid `Directory.GetFiles(path, ...)` when path doesn't exist
  - could copy PathResolver.shade into KoreBuild to avoid updating Sake
  - but, this option can still hide real errors very easily

### DIFF
--- a/build/shade/_k-standard-goals.shade
+++ b/build/shade/_k-standard-goals.shade
@@ -154,13 +154,15 @@ default SAMPLES_PROJECT_GLOB = "samples/*/*.csproj"
 
       projectGlobs.AddRange(srcProjects);
     }
+
     if (!BuildSrcOnly && Directory.Exists("test"))
     {
-        projectGlobs.AddRange(Files.Include(TEST_PROJECT_GLOB).ToList());
+      projectGlobs.AddRange(Files.Include(TEST_PROJECT_GLOB).ToList());
     }
-    if (!BuildSrcOnly && Directory.Exists("samples") && Directory.EnumerateFiles("samples", "*.csproj", SearchOption.AllDirectories).Any())
+
+    if (!BuildSrcOnly && Directory.Exists("samples"))
     {
-        projectGlobs.AddRange(Files.Include(SAMPLES_PROJECT_GLOB).ToList());
+      projectGlobs.AddRange(Files.Include(SAMPLES_PROJECT_GLOB).ToList());
     }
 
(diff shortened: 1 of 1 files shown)


dougbu:
@natemcmaster and @pranavkm this fix is useful for aspnet/Performance#206. My builds currently leave many projects (those only built because they're referenced from a test) in a half-baked state.

build/shade/_k-standard-goals.shade
```diff
@@ -154,13 +154,15 @@ default SAMPLES_PROJECT_GLOB = "samples/*/*.csproj"
...
     }
+
     if (!BuildSrcOnly && Directory.Exists("test"))
     {
-        projectGlobs.AddRange(Files.Include(TEST_PROJECT_GLOB).ToList());
+      projectGlobs.AddRange(Files.Include(TEST_PROJECT_GLOB).ToList());
     }
-    if (!BuildSrcOnly && Directory.Exists("samples") && Directory.EnumerateFiles("samples", "*.csproj", SearchOption.AllDirectories).Any())
+
+    if (!BuildSrcOnly && Directory.Exists("samples"))
```
dougbu, line 163: Put another way, alright if we require an empty samples folder to enable `SAMPLES_PROJECT_GLOB`. But requiring a (no-op) project is a bit much.

build/shade/_k-standard-goals.shade
```diff
@@ -154,13 +154,15 @@ default SAMPLES_PROJECT_GLOB = "samples/*/*.csproj"
...
     {
-        projectGlobs.AddRange(Files.Include(SAMPLES_PROJECT_GLOB).ToList());
+      projectGlobs.AddRange(Files.Include(SAMPLES_PROJECT_GLOB).ToList());
```
dougbu, line 165: Note this line will still throw if `SAMPLES_PROJECT_GLOB` contains an entry that starts with a non-existent folder. One would hope those overriding this variable know what they're doing&hellip;

natemcmaster (approved):

PR #169, opened 2017-02-14
Upgrade CLI version to 1.0.0-rc4-004828

natemcmaster:
This one is going to require manual changes to most of our repos.

For example, NetCoreAppImplicitPackageVersion is deprecated. We used this in every common.props.


pranavkm:
> NetCoreAppImplicitPackageVersion is not deprecated.

Did you mean "now" deprecated?

pranavkm:
I suppose we can remove https://github.com/aspnet/BuildTools/blob/dev/src/Internal.AspNetCore.Sdk/build/Workarounds.targets#L29 from our workarounds if we switch to using `RuntimeFrameworkVersion`

natemcmaster:
Yes, I mean "now" not not. 
I'm testing the removal of the workarounds file.

PR #172, opened 2017-02-15
Patch update 1.1.2

JunTaoLuo:
We can use the file as the primary source of runtime version number but override it with env vars if they are set.


natemcmaster (approved):

JunTaoLuo:
ping @pranavkm @natemcmaster 

I updated the install script to download the runtime from the correct release/1.1.0 channel.

natemcmaster (approved):

pranavkm:
:shipit:

PR #175, opened 2017-02-16
Ignore stdout from sake to determine MSBuild errors

natemcmaster:


BrennanConroy:
So output still goes to console it just ignores it for the purpose of errors and such?

natemcmaster:
Yes. It doesn't try to parse the text of stdout to determine if there were errors. The exit code will determine that.

BrennanConroy:
:shipit:

Issue #179, opened 2017-02-22
Build.cmd truncates double-quoted arguments containing a semicolon


natemcmaster:
**Repro:**
`build.cmd /t:"Compile;Package" /v:d`
or `build.cmd "/t:Compile;Package" /v:d`

build.ps1 receives:
```ps1
$args = ('/t:Compile')
```

Should receive
```ps1
$args = ('/t:Compile;Package', '/v:d')
```

**Workaround:**
Use single quotes
`build.cmd '/t:Compile;Package' /v:d`

natemcmaster:
~~Even worse....it will combine multiple, single quoted args:
`build.cmd '/t:Compile;Package' '/p:Prop=A b'` => `$args = ('/t:Compile;Package /p:Prop=A b')`~~

natemcmaster:
FYI the double-quoted issue remains, but the issue mentioned in my previous argument was resolved by updating the build.ps1 script to use splatting correctly.

dougbu:
@natemcmaster coming back to this issue, I'm wondering if it really matters. Thoughts?

Those concerned about careful quoting can drop into PowerShell and invoke `build.ps1` directly. Or they can double-up their escape characters when using `build.cmd` e.g. using single quotes (as you demonstrated) or `^`. All sound better than complicating `build.cmd`.

FYI, though I've written some ugly re-escaping code in the past, I found nothing more complicated than `%*` in my current `%Path%`.

natemcmaster:
Yeah, we fixed most of this by changing `build.ps1 $args` to `build.ps1 @args`. I'm going to close because more work on this is likely not worth the effort. Re-open if you still think it's a high enough priority.

PR #180, opened 2017-02-22
Make MSBuild the default lifecycle

### LINKED ISSUE #173: Remove NPM, Bower, and Grunt from default lifecycle
Most repos don't need this stuff. Of ~50 repos, only a few need it: SignalR, Performance, Entropy, Diagnostics, plus a few samples in Mvc.

natemcmaster:
Changes the default lifecycle from Sake to MSBuild. This will require updating the command-line syntax for anyone calling 'build.cmd'

Removes NPM, Grunt, and bower from the default lifecycle. Resolves #173 
Replaces the "update volatile feeds" step with a MSBuild target. See https://github.com/aspnet/BuildTools/pull/179
Only calls "restore" once. Lifecycle doesn't override restore.

TODO: 

 - Update Universe build to use new command line syntax
 - Double check repos using makefile.shade. Any repos that still need a makefile.shade can add an MSBuild target that executes Sake or port to MSBuild.


build/KoreBuild.ps1
```diff
@@ -117,14 +117,14 @@ $msBuildArguments = @"
...
 "@
 
-
 if (!(Test-Path $msbuildArtifactsDir))
 {
     mkdir $msbuildArtifactsDir | Out-Null
 }
+
 $msBuildArguments | Out-File -Encoding ASCII -FilePath $msBuildResponseFile
+$allparams        | Out-File -Encoding ASCII -FilePath $msBuildResponseFile -Append
```
pranavkm: Couldn't we move this in to the declaration of `$msBuildArguments`?

build/KoreBuild.sh
```diff
@@ -150,7 +146,7 @@ cat > $msbuildResponseFile <<ENDMSBUILDARGS
 /p:RepositoryRoot="$repoFolder/"
 /fl
 /flp:LogFile="$msbuildLogFile";Verbosity=diagnostic;Encoding=UTF-8
-/p:SakeTargets=$targets
 ENDMSBUILDARGS
+echo -e "$msbuild_args" >> $msbuildResponseFile
```
pranavkm: Same thing - does moving it inline work?

build/targets/KoreBuild.Common.targets
```diff
@@ -53,6 +37,28 @@ as they are subject to change.
 
 <!--
 ###################################################################
+Targets: UseVolatileFeed
+
+Replaces $(RepositoryRoot)\NuGet.config feeds.
+###################################################################
+-->
+  <ItemGroup>
+    <UpdateFeeds Include="AspNetCore" Value="$(NUGET_VOLATILE_FEED_AspNetCore)" Condition="'$(NUGET_VOLATILE_FEED_AspNetCore)' != ''" />
```
pranavkm: This is not the only feed we modify (although it's the one we most commonly change). Should be able to call `Environment.GetEnvironmentVariables` + StartsWith to filter

build/targets/KoreBuild.DefaultLifecycle.targets
```diff
@@ -0,0 +1,26 @@
...
+Standard lifecycle targets.
+
+When extending the solution build, chain off one of these.
+
+The 'DependsOnTargets' are defined in KoreBuild.Common.targets
+###################################################################
+-->
+  <Target Name="PrepareForBuild" DependsOnTargets="CleanArtifacts;CleanSolutions" />
+
+  <Target Name="Initialize" DependsOnTargets="RestoreSolutions" />
```
pranavkm: What of npm, grunt \ gulp?

pranavkm (changes requested):

build/targets/KoreBuild.DefaultLifecycle.targets
natemcmaster: Intentionally removed. As far as I can tell, no one uses grunt, we never supported gulp, and SignalR is the only one that actually builds JavaScript...and it doesn't rely on KoreBuild to do that. The other usages of JS stuff are for samples.

cref #173

build/KoreBuild.sh
natemcmaster: Inlining with the previous statement produced `\n` literals instead of newline characters :-/

build/KoreBuild.ps1
natemcmaster: Yeah probably.

build/targets/KoreBuild.Common.targets
natemcmaster: Which other feeds do we tweak? Can we just add an explicit list?

build/targets/KoreBuild.DefaultLifecycle.targets
```diff
@@ -0,0 +1,24 @@
+<Project>
+
+<!--
+###################################################################
+Standard lifecycle targets.
+
+When extending the solution build, chain off one of these.
+'Restore' is defined in Microsoft.Common.targets
```
natemcmaster: /cc @pranavkm . cref https://github.com/aspnet/KoreBuild/pull/181#issuecomment-282369303

build/targets/KoreBuild.Common.targets
pranavkm: We add one for Universe `NUGET_VOLATILE_FEED_ARTIFACTS` (which is when it's trying to build against each other). We could just use an explicit list for now.

natemcmaster: Thanks. I'd prefer that for simplicity. While possible to use GetEnvironmentVariables, the syntax to split a list of Key=Value pairs is hideous in MSBuild.

natemcmaster:
:up: :date:

davidfowl:
build.cmd still works right?

natemcmaster:
@davidfowl yes, it still works. But it will change the syntax of arguments passed to build.cmd. Instead of `build.cmd target1 target2`, it is now build.cmd + MSBuild args, or `build.cmd /t:Target1 /t:Target2` 

We'll need to make one tweak to build.ps1 to address #179 (changes `korebuild.ps1 $args` to `korebuild.ps1 @args`).

natemcmaster:
:bell:

analogrelay (approved):

natemcmaster:
Merged https://github.com/aspnet/KoreBuild/commit/afa9dcacc2ae0609ad597084756fd28f7d87e469

Issue #177, opened 2017-02-21
"Restore" target in standard-lifecycle overrides the Restore target from Microsoft.Common.targets


pranavkm:
Run `dotnet restore makefile.proj /p:LifecycleType=Standard`.

Expected:
We restore the packages listed in makefile.proj.

Actual:
The `Restore` target is invoked which does not include the makefile.proj. Consequently build tools like `NuGetPackageVerifier` aren't part of restore output.

natemcmaster:
Rename our standard target to "Initialize"?

pranavkm:
Yup, that's what I did. Alternatively we could prefix our targets so it's less ambiguous (although I suspect this isn't a common MSBuildism).

PR #181, opened 2017-02-24
Use a different target name for Restore

pranavkm:
Fixes https://github.com/aspnet/KoreBuild/issues/177

Unblocks https://github.com/aspnet/Coherence/pull/169

### DIFF
--- a/build/targets/standard-lifecycle.targets
+++ b/build/targets/standard-lifecycle.targets
@@ -12,7 +12,7 @@ When extending the solution build, chain off one of these
 
   <Target Name="Clean" DependsOnTargets="CleanArtifacts;CleanSolutions"/>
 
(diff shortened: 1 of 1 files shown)


natemcmaster:
FYI, the alternate solution I have proposed in https://github.com/aspnet/KoreBuild/pull/180 is that we keep "Restore", but that don't override it in the standard lifecycle.

pranavkm:
That ends up including quite a bit more changes to make `_LoadRestoreGraphEntryPoints` work. Let's do this since it's small. Once we merge your changes in, it should continue working without any changes.

natemcmaster:
Ok. :shipit:

natemcmaster:
Yo, one more thing to fix. https://github.com/aspnet/KoreBuild/blob/feature/msbuild/build/shade/_k-standard-goals.shade#L88

PR #185, opened 2017-03-02
Fix cross-plat builds

dougbu:
- need `$(NetFxVersion)` in preflight build; no reference assemblies installed otherwise

nits:
- use a response file for preflight since command is getting long
- set `$(RepositoryRoot)` in preflight, enabling `-r` to override everything
- never use `$makeFileProj` without quotes

Now see expected `Installing NETFrameworkReferenceAssemblies 4.6.1.` output
example error without this:
```
.../Microsoft.Common.CurrentVersion.targets(1111,5): error MSB3644: The reference assemblies for framework ".NETFramework,Version=v4.5.1" were not found. To resolve this, install the SDK or Targeting Pack for this framework version or retarget your application to a version of the framework for which you have the SDK or Targeting Pack installed. Note that assemblies will be resolved from the Global Assembly Cache (GAC) and will be used in place of reference assemblies. Therefore your assembly may not be correctly targeted for the framework you intend. [.../Microsoft.AspNetCore.Mvc.csproj]
```

### DIFF
--- a/build/KoreBuild.sh
+++ b/build/KoreBuild.sh
@@ -119,6 +119,7 @@ fi
 
 makeFileProj="$koreBuildFolder/KoreBuild.proj"
 msbuildArtifactsDir="$repoFolder/artifacts/msbuild"
+msbuildPreflightResponseFile="$msbuildArtifactsDir/msbuild.preflight.rsp"
 msbuildResponseFile="$msbuildArtifactsDir/msbuild.rsp"
 msbuildLogFile="$msbuildArtifactsDir/msbuild.log"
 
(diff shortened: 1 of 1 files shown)


dougbu:
Will merge shortly unless someone objects quickly. Travis builds are solidly red because they always start with an empty packages folder.

pranavkm (approved):

dougbu:
8f2c36b

build/KoreBuild.sh
```diff
@@ -119,24 +119,34 @@ fi
...
 /p:RepositoryRoot="$repoFolder/"
-/p:NetFxVersion=$netfxversion
 /fl
 /flp:LogFile="$msbuildLogFile";Verbosity=detailed;Encoding=UTF-8
 /clp:Summary
+"$makeFileProj"
 ENDMSBUILDARGS
 echo -e "$msbuild_args" >> $msbuildResponseFile
 
-__exec dotnet msbuild /nologo /t:Restore /p:PreflightRestore=true $makeFileProj
```
natemcmaster, line 141: A simpler fix would have been to simply add `/p:NetFxVersion=$netfxversion` here

PR #186, opened 2017-03-02
Restore standalone `/t:Restore`

dougbu:
- move `@(_SolutionItems)` check _inside_ the `_GetRestoreItemsForSolutions` target
  - `DependsOnTargets` are not run before `Condition` is evaluated

### DIFF
--- a/build/targets/KoreBuild.SolutionBuild.targets
+++ b/build/targets/KoreBuild.SolutionBuild.targets
@@ -146,11 +146,12 @@ solutions as well as the current file
 ###################################################################
 -->
   <Target Name="_GetRestoreItemsForSolutions"
-    DependsOnTargets="_ResolveSolutions"
+    DependsOnTargets="_ResolveSolutions;_InnerGetRestoreItemsForSolutions"
     BeforeTargets="_LoadRestoreGraphEntryPoints"
-    Condition=" '$(PreflightRestore)' != 'true' ">
-    <Error Text="No solutions found to build in '$(RepositoryRoot)'" Condition=" '@(_SolutionItems)' == '' " />
+    Condition=" '$(PreflightRestore)' != 'true' " />
 
+  <!-- _GetRestoreItemsForSolutions ensures @(_SolutionItems) has been set up. -->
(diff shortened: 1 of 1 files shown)


dougbu:
At present, `.\build.cmd /t:restore` does nothing in the second `dotnet msbuild` command.

build/targets/KoreBuild.SolutionBuild.targets
```diff
@@ -148,7 +148,9 @@ solutions as well as the current file
   <Target Name="_GetRestoreItemsForSolutions"
     DependsOnTargets="_ResolveSolutions"
     BeforeTargets="_LoadRestoreGraphEntryPoints"
-    Condition="'@(_SolutionItems)' != '' AND '$(PreflightRestore)' != 'true'">
+    Condition=" '$(PreflightRestore)' != 'true' ">
+    <Error Text="No solutions found to build in '$(RepositoryRoot)'" Condition=" '@(_SolutionItems)' == '' " />
```
natemcmaster: We have valid scenarios for running restore without solution items. Can you just skip the MSBuild step instead if _SolutionItems is empty?

natemcmaster (changes requested):

build/targets/KoreBuild.SolutionBuild.targets
dougbu: What are the valid scenarios where `'$(PreflightRestore)' != 'true'`? And which MSBuild step do you mean?

What I'm trying to avoid is having to specify `.\build.cmd "/t:_ResolveSolutions;Restore". That and other target mixes that ensure `@(_SolutionItems)` has been set are what works today.

natemcmaster: Non-standard repo layouts, like our builds that produce package caches and run coherence.

dougbu: :ok: I'll create a second target this one depends on.

dougbu:
:up::date:

natemcmaster (approved):

dougbu:
76998d2

PR #187, opened 2017-03-02
Stop producing TRX files on CI

natemcmaster:
Latest builds of dotnet-test and xunit make producing TRX files unnecessary on TeamCity. In fact, it actually means that TeamCity reports twice as many tests.

### DIFF
--- a/build/targets/KoreBuild.SolutionBuild.targets
+++ b/build/targets/KoreBuild.SolutionBuild.targets
@@ -234,7 +234,6 @@ Runs the VSTest on all projects in the ProjectsToTest itemgroup.
 -->
 
   <PropertyGroup>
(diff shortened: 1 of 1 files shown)


build/targets/KoreBuild.SolutionBuild.targets
```diff
@@ -275,7 +276,7 @@ repository root.
     <NuGetVerifierRuleFile>$(RepositoryRoot)NuGetPackageVerifier.json</NuGetVerifierRuleFile>
   </PropertyGroup>
 
-  <Target Name="VerifyPackages">
+  <Target Name="VerifyPackages" Condition="Exists('$(NuGetVerifierRuleFile)')">
```
dougbu, line 279: Is this just to avoid useless warnings?

dougbu (approved):
Looks good.

build/targets/KoreBuild.SolutionBuild.targets
natemcmaster: Yes.

PR #191, opened 2017-03-05
[question] How do you stop create *.symbols.nupkg

ctaggart:
How do I stop creating symbols packages? We are looking at switch to enable source link. https://github.com/aspnet/Universe/issues/131

### DIFF
--- a/build/targets/KoreBuild.SolutionBuild.targets
+++ b/build/targets/KoreBuild.SolutionBuild.targets
@@ -200,30 +200,6 @@ Executes /t:Pack on all projects matching src/*/*.csproj.
 
 <!--
 ###################################################################
-Target: PackSharedSources
-
-Creates a content files package for all each directory in
-that matches "$(RepositoryRoot)/shared/*.Sources".
-###################################################################
--->
-  <Target Name="PackSharedSources" Condition="Exists('$(SharedSourcesFolder)') OR '@(SharedSourceDirectories)' != ''">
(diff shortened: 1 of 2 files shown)


pranavkm:
It's part of our SDK: https://github.com/aspnet/BuildTools/blob/dev/src/Internal.AspNetCore.Sdk/targets/Common.props#L12.

build/targets/KoreBuild.SolutionBuild.targets
```diff
@@ -200,30 +200,6 @@ Executes /t:Pack on all projects matching src/*/*.csproj.
 
 <!--
 ###################################################################
-Target: PackSharedSources
-
-Creates a content files package for all each directory in
-that matches "$(RepositoryRoot)/shared/*.Sources".
-###################################################################
--->
-  <Target Name="PackSharedSources" Condition="Exists('$(SharedSourcesFolder)') OR '@(SharedSourceDirectories)' != ''">
```
pranavkm, line 209: Not sure I follow - these are for shared src (packages containing src as Content). I don't think source link is meant to be replacement for this.

ctaggart: @pranavkm Right. I misunderstood what this was. Your link above looks to be what I was after. Although, I'm curious if removing this step speeds up the builds any, I probably won't touch https://github.com/aspnet/Mvc/pull/5905 any more until it is reviewed.

PR #194, opened 2017-03-06
Replace resx target with MSBuild

natemcmaster:
This is the second-to-last thing still in Sake.

Companion piece: https://github.com/aspnet/BuildTools/pull/189


natemcmaster:
Here is what `build /t:Resx` would look like on the Localization repo.

image

pranavkm (approved):

PR #198, opened 2017-03-10
Re-enable TRX file production

dougbu:
- reverts part of "Stop producing TRX files on CI"
  - commit e9571513450cb55895680dd2b36a037c3a5bb103
  - leaves bit about avoiding warnings in the `VerifyPackages` target alone


dougbu:
Will update the CI configuration once this goes in. Note we're not sure this will improve reporting when tests fail on a single TFM. If it doesn't, will pull this back out.

natemcmaster (approved):

dougbu:
ffc4f4a

PR #200, opened 2017-03-14
Converting to NETCoreApp2.0

kichalla:


kichalla:
Oops sent too soon. Will update the korebuild.sh file too.

build/KoreBuild.ps1
```diff
@@ -69,6 +69,21 @@ if (!($env:Path.Split(';') -icontains $dotnetLocalInstallFolder))
     $env:Path = "$newPath"
 }
 
+# Install shared runtimes
+if ($sharedRuntimes)
```
natemcmaster: Looks like we need to start using a better file format. JSON?

build/shared-runtime.version
```diff
@@ -1 +1 @@
-1.2.0-beta-001304-00
\ No newline at end of file
+1.1.1;release/1.1.0,2.0.0-beta-001747-00;master
```
natemcmaster: In theory, the 2.0.0 runtime should be bundled with the CLI installer. This file is only for installing runtimes **not** included with the CLI. @pranavkm how do we ensure the package version of Microsoft.NETCore.App on our mirror lines up with the CLI's version?

kichalla: I am a bit unclear about this mismatching NETCoreApp versions. I ran into this issue when doing `dotnet test` when it warned about the `2.0.0-beta-001747-00` version not available. This means that the test runner tool depends on this version and hence the error, is it?? If the mirror were to verify what would we be trying here? that the tools depend on the CLI's netcoreapp versions only?

natemcmaster: The problem comes from mismatch of package and runtime installation. The latest version of the [Microsoft.NETCore.App package](https://dotnet.myget.org/feed/dotnet-core/package/nuget/Microsoft.NETCore.App) on the dotnet-core feed is 2.0.0-beta-001753-00, but the [latest build of .NET Core CLI](https://github.com/dotnet/cli#installers-and-binaries) bundles a different **runtime**, `shared/Microsoft.NETCore.App/2.0.0-beta-001745-00`.

We could make the mirror step smart enough to ensure the package and the runtime match.

Or we can install both. But this will add an extra 45 MB download per build.

pranavkm: The mirroring does try to figure out the shared runtime based on the CLI (https://github.com/aspnet/CoreCLR/blob/dev/tools/MirrorLatestCoreCLR/Program.cs#L97), but we stopped using this because Core-Setup wanted us to not rely on them to get the latest shared runtime. Would be much better to install the shared runtime separately.

pranavkm: The file is intentionally simple because we want it updated from another repo. Let's just install required dependencies i.e. `1.1.1` explicitly and use this file to specify the shared runtime that the gate branch tried out.

build/cli.version
```diff
@@ -1 +1 @@
-1.0.0
+2.0.0-preview1-005418
```
pranavkm, line 1: Does this build fine on the 3 test platforms? We should fix the mirroring code to auto update this rather than managing this manually.

pranavkm (changes requested):

natemcmaster (approved):

natemcmaster:
Before you merge, run a full build on aspnet/Universe to make sure everything compiles.

pranavkm (approved):
#yolo

PR #204, opened 2017-03-21
Assert that OSX >= 10.12 on install

natemcmaster:
.NET Core 2.0 requires macOS Sierra. This adds a check so KoreBuild will fail early if the OS is too old.


BrennanConroy (approved):

build/KoreBuild.sh
```diff
@@ -118,6 +118,15 @@ fi
 
 if [ "$(uname)" == "Darwin" ]; then
     ulimit -n 2048
+
+    # Check that OS is 10.12 or newer
+    osx_version="$(sw_vers | grep ProductVersion | awk '{print $2}')"
```
moozzyk, line 123: Can't you just grep against something like `ProductVersion +10\.1[2-9](\.d+)?` - if found you are good otherwise you exit ?

natemcmaster: It could, but then I'd have to add a second call to sw_vers and grep to fetch the version for the error message

natemcmaster:
:watch: this is going to break most of our OSX builds. Before we do this, I'm going to update our travis.yml files to use mac OS Sierra.

natemcmaster:
Merging now that Travis agents have been updated and internal agents upgrade is in-progress.

PR #220, opened 2017-04-04
Automatically find build number from CI environment

natemcmaster:
One less variable we need to define in our CI config...

### DIFF
--- a/build/targets/KoreBuild.DefaultBuildSettings.props
+++ b/build/targets/KoreBuild.DefaultBuildSettings.props
@@ -1,4 +1,13 @@
 <Project>
+  <!-- Use build number from CI if available -->
+  <PropertyGroup Condition=" '$(BuildNumber)' == '' ">
+    <!--
+      See https://confluence.jetbrains.com/display/TCD10/Predefined+Build+Parameters.
+      This environment variable is automatically set by TeamCity.
+    -->
+    <BuildNumber Condition=" '$(BUILD_NUMBER)' != '' ">$(BUILD_NUMBER)</BuildNumber>
+  </PropertyGroup>
+
   <!-- Create a temporary build number if not assigned by the CI server -->
   <PropertyGroup Condition=" '$(BuildNumber)' == '' ">
     <!--


build/targets/KoreBuild.DefaultBuildSettings.props
```diff
@@ -1,4 +1,16 @@
 <Project>
+  <!-- pickup build number from CI if available -->
+  <PropertyGroup Condition=" '$(BuildNumber)' == '' ">
+    <!--
+      TeamCity=https://confluence.jetbrains.com/display/TCD10/Predefined+Build+Parameters
+      AppVeyor=https://www.appveyor.com/docs/environment-variables/
+      Travis=https://docs.travis-ci.com/user/environment-variables/
+    -->
+    <BuildNumber Condition=" '$(BUILD_NUMBER)' != '' ">$(BUILD_NUMBER)</BuildNumber>
```
pranavkm: Could you make sure that we're setting `BuildCounter` from a previous dependency everywhere we need to and not setting `BuildNumber` to `dep.PreviousConfig.BuildNumber`?

build/targets/KoreBuild.DefaultBuildSettings.props
```diff
@@ -1,4 +1,16 @@
...
+    <BuildNumber Condition=" '$(APPVEYOR_BUILD_NUMBER)' != '' ">$(APPVEYOR_BUILD_NUMBER)</BuildNumber>
```
pranavkm: We super don't care about these builds having a build number.

build/targets/KoreBuild.DefaultBuildSettings.props
natemcmaster: BuildCounter? That's not an MSBuild property we use anywhere. Did you mean BuildNumber?

On UniverseCoherence, I manually set env.BuildNumber = %build.number%, but that's the only place I remember adding it. Most CI configs end up producing packages with temporary build numbers because the BuildNumber is unset.

natemcmaster: FYI: env.BUILD_NUMBER is automatically set by TC to whatever that config's build.number setting is. build.number is not necessarily equal to build.counter.

build/targets/KoreBuild.DefaultBuildSettings.props
natemcmaster: (shrug) It's a "nice to have" that may be useful one day. But if you really want, I'll remove it. It doesn't make a difference at the moment.

build/targets/KoreBuild.DefaultBuildSettings.props
pranavkm: If we have this in any of our build chains:

1. A => `Build_Number` => `build.counter`
2. B => `Build_Number` => `A.Build_Number`

Removing `Build_Number` would no longer be correct. Just wanted you to ensure we weren't doing this.

build/targets/KoreBuild.DefaultBuildSettings.props
pranavkm: I'd rather we remove it if it's not useful.

pranavkm (approved):

Issue #216, opened 2017-03-28
Universe builds cause warning "Microsoft.Common.props cannot be imported again"


natemcmaster:
Something is causing Microsoft.Common.props to be imported multiple times from KoreBuild.proj.

This warning is new, so something might have changed in MSBuild or Microsoft.NET.Sdk.

Example: http://aspnetci/viewLog.html?buildId=193962&buildTypeId=Lite_UniverseTest&tab=buildLog

```
[15:00:29]C:\BuildAgent\work\33bdfc1cae7b2a38\.dotnet\sdk\2.0.0-preview1-005418\Sdks\Microsoft.NET.Sdk\Sdk\Sdk.props(19,3): warning MSB4011: "C:\BuildAgent\work\33bdfc1cae7b2a38\.dotnet\sdk\2.0.0-preview1-005418\15.0\Microsoft.Common.props" cannot be imported again. It was already imported at "C:\BuildAgent\work\33bdfc1cae7b2a38\.dotnet\sdk\2.0.0-preview1-005418\Microsoft.Common.CurrentVersion.targets (19,3)". This is most likely a build authoring error. This subsequent import will be ignored. [C:\BuildAgent\work\33bdfc1cae7b2a38\.build\KoreBuild.proj]
[15:00:29]C:\BuildAgent\work\33bdfc1cae7b2a38\.dotnet\sdk\2.0.0-preview1-005418\Sdks\Microsoft.NET.Sdk\Sdk\Sdk.targets(40,3): warning MSB4011: "C:\BuildAgent\work\33bdfc1cae7b2a38\.dotnet\sdk\2.0.0-preview1-005418\Microsoft.Common.CurrentVersion.targets" cannot be imported again. It was already imported at "C:\BuildAgent\work\33bdfc1cae7b2a38\.dotnet\sdk\2.0.0-preview1-005418\Microsoft.Common.targets (114,3)". This is most likely a build authoring error. This subsequent import will be ignored. [C:\BuildAgent\work\33bdfc1cae7b2a38\.build\KoreBuild.proj]

```

natemcmaster:
We shouldn't use Microsoft.NET.Sdk props or targets from KoreBuild. These targets may alter the default lifecycle.

[...]

PR #227, opened 2017-04-21
Disable MSB4011 import warning

pakrym:
Right now we have to many of them and the way to fix is not totally clear. I'll file an issue to investigate and re-enable.

Also https://github.com/Microsoft/msbuild/blob/6851538897f5d7b08024a6d8435bc44be5869e53/src/Build/Evaluation/Evaluator.cs#L2498 indicates that MSBuild guys want to add some special treatment regarding this import and Sdk attributes.


pranavkm (approved):

build/KoreBuild.ps1
```diff
@@ -117,6 +117,7 @@ if ("${env:CI}${env:APPVEYOR}${env:TEAMCITY_VERSION}${env:TRAVIS}" -eq "")
```
pranavkm, line 117: Don't check in the zip file

pranavkm:
Can you edit the commit to say this is suppressing the duplicate import warning?

natemcmaster:
We shouldn't import files that have an Sdk attribute.

natemcmaster:
https://github.com/aspnet/KoreBuild/issues/216#issuecomment-295983850

pakrym:
@natemcmaster I'll merge it to make it easier to diagnose Universe/Gate builds. We should the right way when we have time or doing changes in the area.

PR #230, opened 2017-04-21
Disable xunit auto reporters when other test loggers are enabled

natemcmaster:
Right now CI shows duplicate test results

1. From TRX logger.
2. From xunit's TeamCity auto-reporter.

This will disable the auto-reporter in favor of TRX reporter.

**FYI**: this won't take effect until we upgrade to xunit 2.3. cref https://github.com/xunit/visualstudio.xunit/pull/108

Cons:

- TC auto reporter reports tests as they happen but TRX only happens when the test process is done and the .trx file is produced
- For new build configs, test results don't light up in TC unless you add the XML report processing feature. https://confluence.jetbrains.com/display/TCD10/XML+Report+Processing. We already have this enabled on all of our test configs.

Pros:

- Reduced console output.
- In some cases (e.g. parallel test runs), the TC reporter does not properly report tests. See #198 
- TRX reporter shows output that TC auto reporter does not. Example:  image

### DIFF
--- a/build/targets/KoreBuild.SolutionBuild.targets
+++ b/build/targets/KoreBuild.SolutionBuild.targets
@@ -289,7 +289,9 @@ Runs the VSTest on all projects in the ProjectsToTest itemgroup.
 -->
 
   <PropertyGroup>
-    <VSTestLogger Condition=" '$(VSTestLogger)' == '' AND '$(TEAMCITY_VERSION)' != '' ">trx</VSTestLogger>
+    <VSTestLogger Condition=" '$(VSTestLogger)' == '' AND '$(TEAMCITY_PROJECT_NAME)' != '' ">trx</VSTestLogger>
+    <!-- ensure AppVeyor's auto test reporter remaines enabled -->
+    <VSTestAutoReporters Condition="'$(APPVEYOR_API_URL)' != ''">true</VSTestAutoReporters>
     <IgnoreFailingTestProjects>false</IgnoreFailingTestProjects>
     <IgnoreFailingTestProjects Condition="'$(KOREBUILD_IGNORE_DOTNET_TEST_EXIT_CODE)' == '1'">true</IgnoreFailingTestProjects>
   </PropertyGroup>
@@ -304,7 +306,9 @@ Runs the VSTest on all projects in the ProjectsToTest itemgroup.
       Condition="'@(_TestProjectItems)' != ''" />
 
     <PropertyGroup>
-    <VSTestNoBuild Condition="'$(VSTestNoBuild)' == ''">$(_SolutionWasBuilt)</VSTestNoBuild>
+      <!-- diable other test reporters if trx logging is enabled -->
+      <TestRunSettings Condition=" '$(VSTestLogger)' != '' AND '$(VSTestAutoReporters)' != 'true' ">RunConfiguration.NoAutoReporters=true</TestRunSettings>
+      <VSTestNoBuild Condition="'$(VSTestNoBuild)' == ''">$(_SolutionWasBuilt)</VSTestNoBuild>
       <_TestContinueOnError Condition="'$(IgnoreFailingTestProjects)' == 'true'">ErrorAndContinue</_TestContinueOnError>
       <_TestContinueOnError Condition="'$(IgnoreFailingTestProjects)' != 'true'">ErrorAndStop</_TestContinueOnError>
     </PropertyGroup>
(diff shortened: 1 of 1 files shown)


dougbu:
Thanks @natemcmaster. I'll look at this tonight.

There's an additional pro: TRX reporter reports some test failures that TC reporter completely ignores. (Our CI reports did _not_ double the number of tests when we re-enabled TRX reporting.)

build/targets/KoreBuild.SolutionBuild.targets
```diff
@@ -289,7 +289,9 @@ Runs the VSTest on all projects in the ProjectsToTest itemgroup.
 -->
 
   <PropertyGroup>
-    <VSTestLogger Condition=" '$(VSTestLogger)' == '' AND '$(TEAMCITY_VERSION)' != '' ">trx</VSTestLogger>
+    <VSTestLogger Condition=" '$(VSTestLogger)' == '' AND '$(TEAMCITY_PROJECT_NAME)' != '' ">trx</VSTestLogger>
```
dougbu: Revert the `TEAMCITY_VERSION` to `TEAMCITY_PROJECT_NAME` change.

build/targets/KoreBuild.SolutionBuild.targets
```diff
@@ -289,7 +289,9 @@ Runs the VSTest on all projects in the ProjectsToTest itemgroup.
...
+    <!-- ensure AppVeyor's auto test reporter remaines enabled -->
```
dougbu: "remains"

build/targets/KoreBuild.SolutionBuild.targets
```diff
@@ -304,15 +306,17 @@ Runs the VSTest on all projects in the ProjectsToTest itemgroup.
       Condition="'@(_TestProjectItems)' != ''" />
 
     <PropertyGroup>
-    <VSTestNoBuild Condition="'$(VSTestNoBuild)' == ''">$(_SolutionWasBuilt)</VSTestNoBuild>
+      <!-- diable other test reporters if trx logging is enabled -->
+      <TestRunSettings Condition=" '$(VSTestLogger)' != '' AND '$(VSTestAutoReporters)' != 'true' ">RunConfiguration.NoAutoReporters=true</TestRunSettings>
```
dougbu: Name this property `VSTestCLIRunSettings`.

dougbu: This should also preserve any existing value e.g.
``` xml
<VSTestCLIRunSettings>$(VSTestCLIRunSettings);RunConfiguration.NoAutoReporters=true</TestRunSettings>
```

build/targets/KoreBuild.SolutionBuild.targets
```diff
@@ -304,15 +306,17 @@ Runs the VSTest on all projects in the ProjectsToTest itemgroup.
...
+      <VSTestNoBuild Condition="'$(VSTestNoBuild)' == ''">$(_SolutionWasBuilt)</VSTestNoBuild>
       <_TestContinueOnError Condition="'$(IgnoreFailingTestProjects)' == 'true'">ErrorAndContinue</_TestContinueOnError>
       <_TestContinueOnError Condition="'$(IgnoreFailingTestProjects)' != 'true'">ErrorAndStop</_TestContinueOnError>
     </PropertyGroup>
 
     <!-- Intentional use of batching ('%') instead of passing items ('@') so that tests fail sooner -->
     <MSBuild Projects="%(_TestProjectItems.Identity)"
       Targets="VSTest"
-      Properties="$(_SolutionProperties);VSTestLogger=$(VSTestLogger);VSTestNoBuild=$(VSTestNoBuild)"
+      Properties="$(_SolutionProperties);VSTestLogger=$(VSTestLogger);VSTestNoBuild=$(VSTestNoBuild);VSTestCLIRunSettings=$(TestRunSettings)"
```
dougbu: Need quotes around the `$(VSTestCLIRunSettings)` value if it can contain semicolons (as I suggested above).

dougbu (changes requested):

build/targets/KoreBuild.SolutionBuild.targets
natemcmaster: Played with it a little. Quotes mess it up, but `.Replace(';','%3B')` should do the trick.

natemcmaster:
:up: :date:

build/targets/KoreBuild.SolutionBuild.targets
```diff
@@ -290,6 +290,8 @@ Runs the VSTest on all projects in the ProjectsToTest itemgroup.
 
   <PropertyGroup>
     <VSTestLogger Condition=" '$(VSTestLogger)' == '' AND '$(TEAMCITY_VERSION)' != '' ">trx</VSTestLogger>
+    <!-- ensure AppVeyor's auto test reporter remains enabled -->
+    <VSTestAutoReporters Condition="'$(APPVEYOR_API_URL)' != ''">true</VSTestAutoReporters>
```
dougbu: Does something set `$(VSTestLogger)` when on AppVeyor? If not, the `$(VSTestLogger)` condition may be enough -- no need for `$(VSTestAutoReporters)`.

dougbu (approved):
:shipit: from me. Agree @pranavkm should have a look too.

build/targets/KoreBuild.SolutionBuild.targets
natemcmaster: I added this as an escape hatch in case we want to force the TeamCity auto-reporter to turn on without needing to make changes KoreBuild.

dougbu: Then support `$(VSTestAutoReporters)` but get rid of the unnecessary special case for `'$(APPVEYOR_API_URL)' != ''` e.g.
``` xml
<VSTestAutoReporters Condition=" '$(VSTestAutoReporters)' == '' AND '$(VSTestLogger)' != '' ">false</VSTestAutoReporters>
<VSTestAutoReporters Condition=" '$(VSTestAutoReporters)' == '' ">true</VSTestAutoReporters>
```

natemcmaster: sounds good

PR #231, opened 2017-04-24
Change default dotnet install dir to USERPROFILE/.dotnet

natemcmaster:
dotnet.exe in .NET Core 2.0 supports multi-level lookup using these rules:

Priority rank below (from 1 to 4):
1. Current working directory
 2. User directory
3. dotnet.exe directory
4. Global .NET directory

See [fx_muxer.cpp](https://github.com/dotnet/core-setup/blob/da184990929677a78c3936fb0abcfca44149c90d/src/corehost/cli/fxr/fx_muxer.cpp#L433-L439)

This means C:\program files\dotnet\dotnet.exe can find the versions of the SDK and shared FX installed to %USERPROFILE%\.dotnet. Setting this as the default install location via KoreBuild should reduce some friction when using VS.

image

### DIFF
--- a/build/KoreBuild.ps1
+++ b/build/KoreBuild.ps1
@@ -20,6 +20,7 @@ if (!$repoFolder) {
 Write-Host "Building $repoFolder"
 cd $repoFolder
 
+$dotnetArch = 'x64'
 $dotnetVersionFile = $PSScriptRoot + "\cli.version"
 $dotnetChannel = "preview"
 $dotnetVersion = Get-Content $dotnetVersionFile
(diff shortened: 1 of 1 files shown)


pranavkm:
:shipit:

kichalla (approved):

kichalla:
@mikeharder FYI

mikeharder (approved):
I like it.  Also matches install dir used on OSX/Linux.

davidfowl:
Last I checked, there was an issue with the compiler that @eerhardt found, I don't know if it was fixed. Other than that, this is a great change. Now we just need dnvm!

eerhardt:
The issue hasn't been fixed yet - https://github.com/dotnet/cli/issues/6298.

However, you can work around this issue by putting a `dotnet.exe` in the root of the hive - `C:\Users\namc\.dotnet\x64\dotnet.exe`.

natemcmaster:
Ah, good to know. Fortunately, the dotnet-install.ps1 script we are using will drop the muxer into %USERPROFLE%\.dotnet\x64.

PR #233, opened 2017-04-27
Cleanup workarounds

### LINKED ISSUE #193: Remove workaround once Microsoft/msbuild#1792 is fixed
When possible, revert the `KoreBuild.sh` part of 0a4214d. Won't need to `tee` once MsBuild does the **Right Thing:tm:**.

### LINKED ISSUE #214: Add retries when downloading nuget.exe
KoreBuild should add retries when downloading nuget.exe to prevent build breaks.

https://github.com/aspnet/KoreBuild/blob/dev/build/KoreBuild.ps1#L94-L99
https://github.com/aspnet/KoreBuild/blob/dev/build/KoreBuild.sh#L147-L151

### LINKED ISSUE #222: Old 1.2.0 version still appearing in some props files
The old 1.2.0 version appears in some props files:
* https://github.com/aspnet/KoreBuild/blob/dev/template/build/dependencies.props#L11
* https://github.com/aspnet/KoreBuild/blob/dev/template/version.props#L4

@natemcmaster - are these not used? Or are they some global default that's always overridden?

natemcmaster:
- No longer download NuGet.exe. Fixes #214
- Don't generate global.json
- ~~Don't suppress color output on CI.~~
- Add comments to template file. Fixes #222

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,7 +14,8 @@ PublishProfiles/
 *.cache
 *.docstates
 _ReSharper.*
-nuget.exe
+*.exe
+*.dll
 *net45.csproj
 *net451.csproj
 *k10.csproj
--- a/build/KoreBuild.ps1
+++ b/build/KoreBuild.ps1
@@ -122,5 +122,5 @@ if (!(Test-Path $msbuildArtifactsDir))
 
 $msBuildArguments | Out-File -Encoding ASCII -FilePath $msBuildResponseFile
 
-exec dotnet msbuild /nologo /t:Restore /p:PreflightRestore=true "$makeFileProj"
+exec dotnet msbuild /nologo /t:Restore /p:PreflightRestore=true /p:RepositoryRoot="$repoFolder/" "$makeFileProj"
 exec dotnet msbuild `@"$msBuildResponseFile"
--- a/build/KoreBuild.ps1
+++ b/build/KoreBuild.ps1
@@ -98,30 +98,10 @@ if (!($env:Path.Split(';') -icontains $dotnetLocalInstallFolder))
     $env:Path = "$newPath"
 }
 
-# wokaround for CLI issue: https://github.com/dotnet/cli/issues/2143
-$sharedPath = (Join-Path (Split-Path ((get-command dotnet.exe).Path) -Parent) "shared");
-(Get-ChildItem $sharedPath -Recurse *dotnet.exe) | %{ $_.FullName } | Remove-Item;
-
-# We still nuget because dotnet doesn't have support for pushing packages
-$nugetExePath = Join-Path $PSScriptRoot 'nuget.exe'
-if (!(Test-Path $nugetExePath))
-{
-    Invoke-WebRequest "https://dist.nuget.org/win-x86-commandline/v4.0.0-rc4/NuGet.exe" -OutFile "$PSScriptRoot/nuget.exe"
-}
-
 $makeFileProj = "$PSScriptRoot/KoreBuild.proj"
 $msbuildArtifactsDir = "$repoFolder/artifacts/msbuild"
 $msbuildLogFilePath = "$msbuildArtifactsDir/msbuild.log"
 $msBuildResponseFile = "$msbuildArtifactsDir/msbuild.rsp"
-
-$preflightClpOption='/clp:DisableConsoleColor'
-$msbuildClpOption='/clp:DisableConsoleColor;Summary'
-if ("${env:CI}${env:APPVEYOR}${env:TEAMCITY_VERSION}${env:TRAVIS}" -eq "")
-{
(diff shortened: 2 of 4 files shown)


build/KoreBuild.ps1
```diff
@@ -142,7 +122,5 @@ if (!(Test-Path $msbuildArtifactsDir))
 
 $msBuildArguments | Out-File -Encoding ASCII -FilePath $msBuildResponseFile
 
-# workaround https://github.com/dotnet/core-setup/issues/1664
```
pranavkm, line 145: Did this get fixed? The work item this pointed to isn't closed as yet?

natemcmaster: I tested. Its been fixed. https://github.com/dotnet/core-setup/pull/2055  It just looks like the bug hasn't been closed.

natemcmaster: I tested. Its been fixed. https://github.com/dotnet/core-setup/pull/2055  It just looks like the bug hasn't been closed.

build/KoreBuild.ps1
```diff
@@ -98,38 +98,18 @@ if (!($env:Path.Split(';') -icontains $dotnetLocalInstallFolder))
...
 
-$preflightClpOption='/clp:DisableConsoleColor'
-$msbuildClpOption='/clp:DisableConsoleColor;Summary'
-if ("${env:CI}${env:APPVEYOR}${env:TEAMCITY_VERSION}${env:TRAVIS}" -eq "")
-{
-    # Not on any of the CI machines. Fine to use colors.
-    $preflightClpOption=''
-    $msbuildClpOption='/clp:Summary'
-}
-
```
dougbu: This wasn't just a workaround and should not be removed. None of the CI systems display MSBuild's colours. So, this update will only increase log sizes.

Same for removing this feature on other platforms.

build/KoreBuild.sh
```diff
@@ -133,68 +126,31 @@ else
...
 /p:RepositoryRoot="$repoFolder/"
 /fl
 /flp:LogFile="$msbuildLogFile";Verbosity=detailed;Encoding=UTF-8
-$msbuildClpOption
+/clp:Summary
 "$makeFileProj"
 ENDMSBUILDARGS
 echo -e "$msbuild_args" >> $msbuildResponseFile
 
+__exec dotnet msbuild /nologo /t:Restore /p:PreflightRestore=true /p:NetFxVersion=$netfxversion "$makeFileProj"
```
dougbu: Why doesn't this command set `$(RepositoryRoot)` anymore?

.gitignore
```diff
@@ -14,7 +14,8 @@ PublishProfiles/
 *.cache
 *.docstates
 _ReSharper.*
-nuget.exe
+*.exe
+*.dll
```
dougbu, line 18: What DLLs and EXEs now need to be ignored?

dougbu (changes requested):
Generally, not sure why this cleanup matters much.

build/KoreBuild.sh
natemcmaster: It was never needed for PreflightRestore.

build/KoreBuild.ps1
natemcmaster: We disabled color and did the 'tee' on Linux to workaround  https://github.com/Microsoft/msbuild/issues/1792 which has been fixed since mid March.

MSBuild console logger doesn't emit much color in minimal mode, so I don't think log size is a major concern: on the test run for this PR, it added a total of 76 bytes: https://travis-ci.org/aspnet/KoreBuild/jobs/226232644.

dougbu: I made this change and Microsoft/msbuild#1792 was _part_ of the reason behind it. There's no point to colour in the CI builds.

natemcmaster: >  There's no point to colour in the CI builds.

I disagree. Color is useful, esp. when there are warnings and errors. AppVeyor is the only one that strips color. TeamCity shows color when you view logs in "console view".
image

pranavkm (approved):

natemcmaster:
Ping @dougbu

.gitignore
dougbu: Ping @natemcmaster

build/KoreBuild.ps1
dougbu: Right, we disagree here. Never seen anyone use the console view in TC.

.gitignore
natemcmaster: Nothing new. Just general cleanup.

build/KoreBuild.ps1
natemcmaster: Ok. I'll back our the this particular change and make a separate discussion issue for this.

dougbu (approved):

PR #240, opened 2017-05-10
Install CLI after installing shared runtimes

kichalla:
Currently `dotnet.exe` gets overwritten by the installation of shared runtimes which was the cause of the behavior where when after typing `dotnet` and pressing enter, it crashes. This re-order fixes it.

### DIFF
--- a/build/KoreBuild.sh
+++ b/build/KoreBuild.sh
@@ -104,6 +104,16 @@ else
     export DOTNET_INSTALL_DIR=$DOTNET_INSTALL_DIR
     chmod +x $scriptRoot/dotnet/dotnet-install.sh
 
+    # Temporarily install these runtimes to prevent build breaks for repos not yet converted
+    # 1.0.4 - for tools
+    install_shared_runtime "1.0.4" "preview"
+    # 1.1.1 - for test projects which haven't yet been converted to netcoreapp2.0
+    install_shared_runtime "1.1.1" "release/1.1.0"
+
+    if [ "$sharedRuntimeVersion" != "" ]; then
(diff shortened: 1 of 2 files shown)


build/KoreBuild.sh
```diff
@@ -98,6 +98,16 @@ if [ ! -z "$KOREBUILD_SKIP_RUNTIME_INSTALL" ]; then
     # Add to the _end_ in case preferred .NET CLI is not in the default location.
     [[ ":$PATH:" != *":$DOTNET_INSTALL_DIR:"* ]] && export PATH="$PATH:$DOTNET_INSTALL_DIR"
 else
+    # Temporarily install these runtimes to prevent build breaks for repos not yet converted
```
natemcmaster: These should be moved down few lines, after DOTNET_INSTALL_DIR is set and dotnet-install.sh has been chmod-ed

natemcmaster:
This should work. Another thing you might consider adding is setting `OverwriteNonVersionedFiles` to false when the `-SharedRuntime` flag is passed to dotnet-install.ps1.
https://github.com/aspnet/KoreBuild/blob/dev/build/dotnet/dotnet-install.ps1#L76

kichalla:
> This should work. Another thing you might consider adding is setting OverwriteNonVersionedFiles to false when the -SharedRuntime flag is passed to dotnet-install.ps1.
https://github.com/aspnet/KoreBuild/blob/dev/build/dotnet/dotnet-install.ps1#L76

While it might work, I think we should not rely on it as its an internal detail(and not like let's say an input parameter) and could be changed without notice

natemcmaster (approved):

PR #241, opened 2017-05-12
Use binary logging instead of plaintext file logging

### LINKED ISSUE #225: Use binary logger
https://github.com/Microsoft/MSBuild/wiki/Binary-Log

natemcmaster:
This will save us a huge chunk of disk space in our CI artifacts.

To view the log, open the \*.binlog file with https://github.com/KirillOsenkov/MSBuildStructuredLog

Close #225


mikeharder (approved):
Approved on the condition that the MSBuild console output is still written to the TeamCity build log.

natemcmaster:
Console output is the same. See https://ci.appveyor.com/project/aspnetci/korebuild/build/1.0.90 for example.

PR #243, opened 2017-05-17
add support for converting log to structured log format

analogrelay:
This adds a post-build step to our build scripts to convert the `.binlog` MSBuild Binary log file to a `.buildlog` Structured log file. The resulting file is much smaller, and loads much quicker into the structured log viewer.

Note that we do not remove the `.binlog` file because the `.buildlog` file can't be replayed to a text file, so we need to keep the `.binlog` if we want that functionality.

Also, this is behind an environment variable trigger that I'll enable on the CI. It takes a good 40 seconds to run in Universe, and 10-15 seconds for some larger repos, so I think it's better to have it off by default. Devs can always process the log themselves on-demand.

### DIFF
--- a/build/KoreBuild.ps1
+++ b/build/KoreBuild.ps1
@@ -123,4 +123,20 @@ if (!(Test-Path $msbuildArtifactsDir))
 $msBuildArguments | Out-File -Encoding ASCII -FilePath $msBuildResponseFile
 
 __exec dotnet restore /p:PreflightRestore=true "$makeFileProj"
-__exec dotnet msbuild `@"$msBuildResponseFile"
+
+try {
+    __exec dotnet msbuild `@"$msBuildResponseFile"
+} finally {
+    # If the structured logger is enabled, replay the binary log to a structured log and remove the binary log
+    # (can't do this in MSBuild because MSBuild logs the log, for obvious reasons)
+    if ($env:KOREBUILD_WRITE_STRUCTURED_LOG -eq "true") {
+        # Open to suggestions for a cleaner way to do this...
+        $r = [regex]"!StructuredLoggerVersion:([^!]*)!"
+        $m = $r.Match((dotnet build $makeFileProj /t:GetStructuredLoggerVersion /p:DisableDefaultTargets=true))
+        $StructuredLoggerPath = "$env:USERPROFILE\.nuget\packages\microsoft.build.logging.structuredlogger\$($m.Groups[1].Value)\lib\netstandard1.5\StructuredLogger.dll"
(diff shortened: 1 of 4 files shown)


build/KoreBuild.ps1
```diff
@@ -123,4 +123,20 @@ if (!(Test-Path $msbuildArtifactsDir))
 $msBuildArguments | Out-File -Encoding ASCII -FilePath $msBuildResponseFile
 
 __exec dotnet restore /p:PreflightRestore=true "$makeFileProj"
-__exec dotnet msbuild `@"$msBuildResponseFile"
+
+try {
+    __exec dotnet msbuild `@"$msBuildResponseFile"
+} finally {
+    # If the structured logger is enabled, replay the binary log to a structured log and remove the binary log
```
kichalla, line 130: nit: remove the part `and remove the binary log ` as we want to be able to replay to text file if needed?

build/KoreBuild.ps1
```diff
@@ -123,4 +123,20 @@ if (!(Test-Path $msbuildArtifactsDir))
...
+    # (can't do this in MSBuild because MSBuild logs the log, for obvious reasons)
+    if ($env:KOREBUILD_WRITE_STRUCTURED_LOG -eq "true") {
+        # Open to suggestions for a cleaner way to do this...
+        $r = [regex]"!StructuredLoggerVersion:([^!]*)!"
+        $m = $r.Match((dotnet build $makeFileProj /t:GetStructuredLoggerVersion /p:DisableDefaultTargets=true))
+        $StructuredLoggerPath = "$env:USERPROFILE\.nuget\packages\microsoft.build.logging.structuredlogger\$($m.Groups[1].Value)\lib\netstandard1.5\StructuredLogger.dll"
```
kichalla, line 136: Is this path problematic? like in our CI we have the package folder in the build agent working dir...also could you update the travis and appveyor xml files for this PR so that we can take a look at how this works on them (and hopefully get the artifacts)

kichalla (changes requested):

natemcmaster:
My two cents: since this feature is off by default, putting this in KoreBuild makes it harder to use. I'm unlikely to re-run a whole build just so I can get a `.buildlog` file. I'd rather take the `.binlog` file and convert it to `buildlog` in a separate step so I can avoid a rebuild. IMO it would be better to have a separate script that runs this conversion independent of a KoreBuild run.

But if you keep it in korebuild, can you set KOREBUILD_WRITE_STRUCTURED_LOG in appveyor.yml and travis.yml for this repo so our canary build will run this?

analogrelay:
We can do it in CI steps. I put it here originally so devs _can_ access it, but you're right, they can just process the log themselves.

PR #249, opened 2017-05-25
Updated shared runtimes to install to 1.0.5 & 1.1.2

kichalla:
The latest CLI that the mirror is trying to use 6184, when seeing a netcoreapp1.1 tfm tries to find out a shared fx version of 1.1.2. (not sure if this is expected as shouldn't it fallback to the available 1.1 compatible runtimes?). I need this for the mirror to go through.

### DIFF
--- a/build/KoreBuild.sh
+++ b/build/KoreBuild.sh
@@ -105,9 +105,9 @@ else
     chmod +x $scriptRoot/dotnet/dotnet-install.sh
 
     # Temporarily install these runtimes to prevent build breaks for repos not yet converted
-    # 1.0.4 - for tools
+    # 1.0.5 - for tools
     install_shared_runtime "1.0.4" "preview"
-    # 1.1.1 - for test projects which haven't yet been converted to netcoreapp2.0
+    # 1.1.2 - for test projects which haven't yet been converted to netcoreapp2.0
     install_shared_runtime "1.1.1" "release/1.1.0"
 
     if [ "$sharedRuntimeVersion" != "" ]; then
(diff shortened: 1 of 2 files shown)


mikeharder (approved):

mikeharder:
Merging to test this.  If it doesn't work, maybe we'll need to install all 4 versions (1.0.4, 1.0.5, 1.1.1, 1.1.2).

mikeharder:
@kichalla: Please delete branch when ready

build/KoreBuild.sh
```diff
@@ -105,9 +105,9 @@ else
     chmod +x $scriptRoot/dotnet/dotnet-install.sh
 
     # Temporarily install these runtimes to prevent build breaks for repos not yet converted
-    # 1.0.4 - for tools
+    # 1.0.5 - for tools
     install_shared_runtime "1.0.4" "preview"
-    # 1.1.1 - for test projects which haven't yet been converted to netcoreapp2.0
+    # 1.1.2 - for test projects which haven't yet been converted to netcoreapp2.0
     install_shared_runtime "1.1.1" "release/1.1.0"
```
natemcmaster, line 111: umm, this is still installing 1.1.1 and 1.0.4 on xplat...

natemcmaster:
It's a shame we still need 1.x at all. We should really upgrade our build tools to 2.0

build/KoreBuild.sh
mikeharder: Good catch, I will fix this.

mikeharder:
@natemcmaster: Please open an issue to upgrade build tools to 2.0 if you think it's something we should do.

natemcmaster:
https://github.com/aspnet/BuildTools/issues/200...it's been open for a while :-/

build/KoreBuild.sh
kichalla: Oops. thanks

PR #252, opened 2017-05-26
Added a common property to always use the latest shared runtime

kichalla:
Instead of changing all repos with this property, we could change it here and it seems to work. But I am not sure if this would have any other consequences.


natemcmaster:
Unfortunately this won't be picked up by projects. This has to be added to dependencies.props in every repo.

kichalla:
> Unfortunately this won't be picked up by projects. This has to be added to dependencies.props in every repo.

You are right. Seems like my environment wasn't setup right. As you mentioned, putting it in depedency.props is working. Closing this issue now.

PR #256, opened 2017-06-07
Make binary logging conditional

smitpatel:
It has terrible memory consumption especially on Ubuntu.
Ref: https://github.com/Microsoft/msbuild/issues/2168
It causes failures on Ubuntu https://github.com/Microsoft/msbuild/issues/2164

This PR makes it conditional
Set $env:KOREBUILD_ENABLE_BINARY_LOG=1 to enable logging.

### DIFF
--- a/build/KoreBuild.ps1
+++ b/build/KoreBuild.ps1
@@ -119,15 +119,22 @@ if (!($env:Path.Split(';') -icontains $dotnetLocalInstallFolder))
 
 $makeFileProj = "$PSScriptRoot/KoreBuild.proj"
 $msbuildArtifactsDir = "$repoFolder/artifacts/msbuild"
-$msbuildLogFilePath = "$msbuildArtifactsDir/msbuild.binlog"
 $msBuildResponseFile = "$msbuildArtifactsDir/msbuild.rsp"
 
+$msBuildLogArgument = ""
+
+if ($env:KOREBUILD_ENABLE_BINARY_LOG -eq "1")
+{
+    Write-Host "Enabling binary logging because KOREBUILD_ENABLE_BINARY_LOG = 1"
+    $msbuildLogFilePath = "$msbuildArtifactsDir/msbuild.binlog"
+    $msBuildLogArgument = "/bl:$msbuildLogFilePath"
+}
 
 $msBuildArguments = @"
 /nologo
 /m
 /p:RepositoryRoot="$repoFolder/"
-/bl:"$msbuildLogFilePath"
+"$msBuildLogArgument"
 /clp:Summary
 "$makeFileProj"
 "@
--- a/build/KoreBuild.sh
+++ b/build/KoreBuild.sh
@@ -152,7 +152,13 @@ export ReferenceAssemblyRoot=$NUGET_PACKAGES/netframeworkreferenceassemblies/$ne
 makeFileProj="$scriptRoot/KoreBuild.proj"
 msbuildArtifactsDir="$repoFolder/artifacts/msbuild"
 msbuildResponseFile="$msbuildArtifactsDir/msbuild.rsp"
-msbuildLogFile="$msbuildArtifactsDir/msbuild.binlog"
+msBuildLogArgument=""
+
+if [ ! -z "$KOREBUILD_ENABLE_BINARY_LOG" ]; then
+    echo "Enabling binary logging because KOREBUILD_ENABLE_BINARY_LOG is set"
+    msBuildLogFile="$msbuildArtifactsDir/msbuild.binlog"
+    msBuildLogArgument="/bl:$msBuildLogFile"
+fi
 
 if [ ! -f $msbuildArtifactsDir ]; then
     mkdir -p $msbuildArtifactsDir
@@ -162,7 +168,7 @@ cat > $msbuildResponseFile <<ENDMSBUILDARGS
 /nologo
 /m
 /p:RepositoryRoot="$repoFolder/"
-/bl:"$msbuildLogFile"
+"$msBuildLogArgument"
 /clp:Summary
 "$makeFileProj"
 ENDMSBUILDARGS


KirillOsenkov:
I've posted a workaround here:
https://github.com/Microsoft/msbuild/issues/2168#issuecomment-306937441

Would you mind trying that first?

smitpatel:
@natemcmaster @mikeharder @kichalla - for review.
My PS & Bash skills are not awesome. :trollface: 
cc: @Eilon

build/KoreBuild.sh
```diff
@@ -152,7 +152,13 @@ export ReferenceAssemblyRoot=$NUGET_PACKAGES/netframeworkreferenceassemblies/$ne
 makeFileProj="$scriptRoot/KoreBuild.proj"
 msbuildArtifactsDir="$repoFolder/artifacts/msbuild"
 msbuildResponseFile="$msbuildArtifactsDir/msbuild.rsp"
-msbuildLogFile="$msbuildArtifactsDir/msbuild.binlog"
+msBuildLogArgument=""
+
+if [ ! -z "$KOREBUILD_SKIP_RUNTIME_INSTALL" ]; then
```
natemcmaster: copy/paste mistake? Should be KOREBUILD_ENABLE_BINARY_LOG

natemcmaster:
@KirillOsenkov just curious...how does specifying `/logger:BinaryLogger` vs `/bl` workaround the issue? It may be a little tough for us to do here as the location of the `dotnet/sdk/$version` folder varies from machine to machine and project to project.

KirillOsenkov:
When you use `/bl` this line gets run:
https://github.com/Microsoft/msbuild/blob/master/src/MSBuild/XMake.cs#L2741-L2744

This is actually what is causing your problems, not the binary logger itself. If you remove binary logger, but add `/v:diag` to the build, you'll hit the same perf issues.

Specifying the binary logger via the /logger doesn't take that codepath so it's not hitting the issue.

The real problem is NuGet logging a ridiculous amount of information as inputs to its tasks (2MB string each, thousands of strings). Not sure what can be done there.

KirillOsenkov:
This is a PR for MSBuild to remove the offending line:
https://github.com/Microsoft/msbuild/pull/2193

natemcmaster:
Nvm, just realized `dotnet --info` gives us the location of the sdk. Let's try the workaround: https://github.com/aspnet/KoreBuild/pull/257

mikeharder (approved):

Eilon (approved):

natemcmaster:
:shipit:

nguerrera:
Agree the amount of data passed to NuGet tasks is staggering, but it's exacerbated by msbuild concatenating all items passed to a task parameter in to a single string. Surely these inputs could be logged more efficiently.

build/KoreBuild.ps1
```diff
@@ -119,15 +119,22 @@ if (!($env:Path.Split(';') -icontains $dotnetLocalInstallFolder))
 
 $makeFileProj = "$PSScriptRoot/KoreBuild.proj"
 $msbuildArtifactsDir = "$repoFolder/artifacts/msbuild"
-$msbuildLogFilePath = "$msbuildArtifactsDir/msbuild.binlog"
 $msBuildResponseFile = "$msbuildArtifactsDir/msbuild.rsp"
 
+$msBuildLogArgument = ""
+
+if ($env:KOREBUILD_ENABLE_BINARY_LOG -eq "1")
```
dougbu, line 126: Why `"1"` instead of `"true"`?

build/KoreBuild.sh
```diff
@@ -152,7 +152,13 @@ export ReferenceAssemblyRoot=$NUGET_PACKAGES/netframeworkreferenceassemblies/$ne
 makeFileProj="$scriptRoot/KoreBuild.proj"
 msbuildArtifactsDir="$repoFolder/artifacts/msbuild"
 msbuildResponseFile="$msbuildArtifactsDir/msbuild.rsp"
-msbuildLogFile="$msbuildArtifactsDir/msbuild.binlog"
+msBuildLogArgument=""
+
+if [ ! -z "$KOREBUILD_ENABLE_BINARY_LOG" ]; then
```
dougbu, line 157: Why isn't this the `bash` equivalent of the Windows condition?

build/KoreBuild.ps1
smitpatel: copied the format as `KOREBUILD_SKIP_RUNTIME_INSTALL`

build/KoreBuild.sh
smitpatel: copied the format as `KOREBUILD_SKIP_RUNTIME_INSTALL`

KirillOsenkov:
@nguerrera I've filed a bug https://github.com/Microsoft/msbuild/issues/2200 to investigate unnecessary string concatenation in log task inputs that you mention.

Could you please add more details to it (basically dump your brain about the issue). The more we know the easier it'll be to investigate.

build/KoreBuild.sh
dougbu: `KOREBUILD_SKIP_RUNTIME_INSTALL` is bizarrely inconsistent. The rest of this script uses e.g. `if ($env:KOREBUILD_DOTNET_CHANNEL)` for PowerShell and the matching `[ -z "$KOREBUILD_DOTNET_CHANNEL" ] &&` for `bash`.

build/KoreBuild.ps1
```diff
@@ -119,15 +119,22 @@ if (!($env:Path.Split(';') -icontains $dotnetLocalInstallFolder))
...
+{
+    Write-Host "Enabling binary logging because KOREBUILD_ENABLE_BINARY_LOG = 1"
+    $msbuildLogFilePath = "$msbuildArtifactsDir/msbuild.binlog"
+    $msBuildLogArgument = "/bl:$msbuildLogFilePath"
```
KirillOsenkov, line 130: So sorry I've missed it, but `/bl` will cause the perf problems for you. Didn't we decide to use the `/logger` parameter instead?

build/KoreBuild.sh
dougbu: @smitpatel could you file a follow-up issue to make this and `KOREBUILD_SKIP_RUNTIME_INSTALL` consistent with the other environment variables used in these scripts.

smitpatel: Feel free to file an issue (or even do clean up) as neither I know what is best practice in scripting languages nor I know practice followed in this repo around this.

I created this PR simply to unblock EF on travis.

build/KoreBuild.ps1
smitpatel: This will cause perf issues but we are not using it everywhere anymore. Using `/logger` here would be a work-around (which needs to be removed when msbuild fixes the perf issue). As only small set of cases we are going to generate binary log, perhaps those machines can take the perf impact. If we find a compelling reason that we need work-around, we will update to `/logger`.
Thanks for all the help. 😄

build/KoreBuild.sh
dougbu: #259

PR #260, opened 2017-06-08
Update dotnet-install scripts to match latest from dotnet/cli repo

mikeharder:


build/dotnet/dotnet-install.sh
```diff
@@ -549,10 +549,6 @@ download() {
     local out_path=${2:-}
 
     local failed=false
-    # Restart the request up to N times if it fails
```
mikeharder, line 552: I think we should try reverting these changes and see if it causes issues (e.g. on Travis).  If so, we should submit a PR to https://github.com/dotnet/cli rather than change our copy.

natemcmaster: fine with me

natemcmaster (approved):

kichalla (approved):

mikeharder:
Will merge once AppVeyor and Travis pass, and I've locally verified this script can install old and new CLI versions.

Issue #224, opened 2017-04-13
Add a way to disable package restore when building repos


kichalla:
Sometimes I want to just work with the packages which are already present on my machine. Without this frequent coherence builds cause package restore failures when building a repo.

cc @dougbu

Eilon:
Is the issue more about packages not being coherent, or about not wanting to get any new packages at all?

dougbu:
For me, the issue is about not wanting any new packages. I want to be able to _choose_ when it's time to `git pull` and, even more often, want to build from root when disconnected.

kichalla:
> Is the issue more about packages not being coherent, or about not wanting to get any new packages at all?

yeah the first case because at the time filing of this bug there were frequent successful coherence builds and doing a build locally would fail due to coherent set not available. Earlier there used to be a switch to disable restore if it was already done. We need something like that again.

PR #262, opened 2017-06-09
Added a way to disable package restore when compiling a solution

kichalla:
Usage: `build.cmd /p:NoRestore=true`

Fixes: https://github.com/aspnet/KoreBuild/issues/224

### DIFF
--- a/build/targets/KoreBuild.Common.targets
+++ b/build/targets/KoreBuild.Common.targets
@@ -10,7 +10,8 @@ extending the *DependsOn property
 -->
   <PropertyGroup>
     <PrepareDependsOn></PrepareDependsOn>
-    <CompileDependsOn>Prepare;Restore</CompileDependsOn>
+    <CompileDependsOn>Prepare</CompileDependsOn>
+    <CompileDependsOn Condition="'$(NoRestore)' != 'true'">$(CompileDependsOn);Restore</CompileDependsOn>
     <PackageDependsOn>Compile</PackageDependsOn>
     <TestDependsOn>Package</TestDependsOn>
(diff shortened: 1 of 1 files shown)


smitpatel (approved):

build/targets/KoreBuild.Common.targets
```diff
@@ -10,7 +10,8 @@ extending the *DependsOn property
 -->
   <PropertyGroup>
     <PrepareDependsOn></PrepareDependsOn>
-    <CompileDependsOn>Prepare;Restore</CompileDependsOn>
+    <CompileDependsOn>Prepare</CompileDependsOn>
+    <CompileDependsOn Condition="'$(NoRestore)'=='false' Or '$(NoRestore)'==''">$(CompileDependsOn);Restore</CompileDependsOn>
```
dougbu: This condition should be simpler and as specific as the one for the new `<Warning/>`: `'$(NoRestore)' != 'true'`

dougbu: Also add the spaces around the conditional operator here and in the new condition below.

dougbu (changes requested):

kichalla:
🆙 📅

dougbu (approved):
Thanks!

Will this disable the "preflight" restore as well?

kichalla:
> Will this disable the "preflight" restore as well?

No, it actually does not.

dougbu:
> No, it actually does not.

That's somewhat separate. Please file a follow-up bug about supporting offline builds. (The new property won't work there. Perhaps we should check a `KOREBUILD_NORESTORE` environment variable in a few places?)

PR #265, opened 2017-06-27
Allow source package to have props file

pakrym:
Useful if for example you want to include source file from another location.


natemcmaster (approved):
Can you also port this to here?

https://github.com/aspnet/BuildTools/blob/dev/sdk/KoreBuild/modules/sharedsources/sharedsources.csproj

PR #271, opened 2017-08-18
Install 1.1.2 shared runtime

JunTaoLuo:
I don't know if we need to install both 1.1.1 and 1.1.2 but I think we can start with just 1.1.2 and if it doesn't work, we can add 1.1.1 as well.


pranavkm:
Sure. Let's give this a go.

JunTaoLuo:
Force updated the rel/1.1.2 tag

